PR #16, opened 2024-09-22
Simplify server auto-configuration

onobc:
This simplifies the server-side auto-configuration as follows:

- delete svc discover abstraction (use ObjectProvider for now)
- use single AC class

cc: @dsyer


spring-grpc-core/src/main/java/org/springframework/grpc/server/lifecycle/GrpcServerLifecycle.java
```diff
@@ -107,6 +105,11 @@ public int getPort() {
 		return this.server == null ? 0 : this.server.getPort();
 	}
 
+	@Override
+	public void setApplicationEventPublisher(ApplicationEventPublisher eventPublisher) {
```
onobc, line 109: I believe this is the more Spring idiomatic way.

PR #17, opened 2024-09-22
Simplify server auto-configuration

onobc:
This simplifies the server-side auto-configuration as follows:

- delete svc discover abstraction (use ObjectProvider for now)
- use single AC class

cc: @dsyer


spring-grpc-core/src/main/java/org/springframework/grpc/server/lifecycle/GrpcServerLifecycle.java
```diff
@@ -107,6 +105,11 @@ public int getPort() {
 		return this.server == null ? 0 : this.server.getPort();
 	}
 
+	@Override
+	public void setApplicationEventPublisher(ApplicationEventPublisher eventPublisher) {
```
onobc, line 109: I believe this is the more Spring idiomatic way.

spring-grpc-core/src/test/java/org/springframework/grpc/server/DefaultGrpcServerFactoryTests.java
```diff
@@ -13,14 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.springframework.grpc.autoconfigure.server;
```
onobc, line 16: I have no clue why Github insists this is a move of the deleted GrpcServiceDiscover - but it is not. Please disregard. It is a newly added test placeholder

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
```diff
@@ -17,36 +17,51 @@
...
-@ConditionalOnBean(BindableService.class)
-@AutoConfigureBefore(GrpcServiceAutoConfiguration.class)
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for gRPC server-side components.
+ *
+ * @author David Syer
+ * @author Chris Bono
+ */
+@AutoConfiguration
+@ConditionalOnClass(BindableService.class)
```
onobc, line 41: I believe this was a bug before as `@ConditionalOnBean` would not fire since this is the config that adds the BS beans. Also, we want to disable if they do not have `io.grpc` on the classpath.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
```diff
@@ -17,36 +17,51 @@
...
 	@ConditionalOnMissingBean(GrpcServerFactory.class)
 	@Bean
-	public DefaultGrpcServerFactory<?> defaultGrpcServerFactory(final GrpcServerProperties properties,
-			final GrpcServiceDiscoverer serviceDiscoverer, final List<GrpcServerConfigurer> serverConfigurers) {
-		final DefaultGrpcServerFactory<?> factory = new DefaultGrpcServerFactory<>(properties.getAddress(),
-				properties.getPort(), serverConfigurers);
-		for (final ServerServiceDefinition service : serviceDiscoverer.findGrpcServices()) {
-			factory.addService(service);
-		}
+	DefaultGrpcServerFactory<?> defaultGrpcServerFactory(ObjectProvider<BindableService> grpcServicesProvider,
```
onobc, line 53: I removed the `GrpcServiceDiscoverer` concept in favor of an `ObjectProvider<BindableService>`. We can introduce a discoverer later when needed is my thinking.

spring-grpc-core/src/main/java/org/springframework/grpc/server/lifecycle/GrpcServerLifecycle.java
dsyer: I don't think it matters does it? Don't we generally prefer constructor injection?

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
dsyer: I'm not 100% convinced about `ObjectProvider` (a simple `List` is easier to test), but this seems like a reasonable point generally.

spring-grpc-core/src/main/java/org/springframework/grpc/server/lifecycle/GrpcServerLifecycle.java
onobc: I tend to see `ApplicationContext` and `AEP` passed in via the `<thing>Aware` contract. But, it is probably more of a preference. Constructor injection is a good thing too and make it easier to test. I will omit this from the proposal.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
onobc: Most of the Spring Boot auto-configs are now using OP vs List and it is pretty easy to test w/ auto-configuration tests by setting up BS beans in the context and making sure those come into the server factory.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
onobc: > and the integration tests seem to prove it)

Ahh, I just noticed these under `samples` module. 👍🏻 

> The BS beans are added by marking them as `@Component`

I missed this fact. So basically, if there are no BS beans then there is no need to auto-config the server. Would it ever be the case that one would want to add a service after startup? If so, we may want to remove this constraint. If we did, then the object provider would just return zero services and the server would start but not provide any services at that point.

> Why would you be using spring-grpc without io.grpc on the classpath?

The scenario is when this later lives in Spring Boot AC module it will be necesarry.
I am approaching the auto-configuration (AC) as if it lives in Spring Boot because:

1. I expect that is where we want it long term
2. the more it looks like that then it will be easier to make the move

I did this for spring-pulsar and it made the move into SB straightforward.

spring-grpc-core/src/main/java/org/springframework/grpc/server/lifecycle/GrpcServerLifecycle.java
onobc: Done

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
onobc: > and the integration tests seem to prove it

I believe the ITs are passing.

dsyer: That's not the point. The point is that they were passing before, so you can put the `@ConditionalOnBean` back in.

onobc: Done. I also added an auto-configuration test for the server config.

PR #23, opened 2024-10-05
Consolidate samples into single integration test

onobc:
Moves the current sample variants into a single integration test (GrpcServerApplicationTests) leveraging the Junit5 `@Nested` feature.

### Motivation
I was going to add more "samples" in different sample modules to restrict the classpath (e.g. shaded only, okhttp provider, ..) and in preparation for that I consolidated the current tests into a single test.

I then came to the conclusion that the above tests I was thinking of adding were edge-casey and we already have enough test coverage for them in the regular core/autoconfigure tests.

I like the tests in this consolidated form as they are really "GrpcServerFactory" related and now in a single file. Future sample tests will be easier to group/identify this way.


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
```diff
@@ -24,7 +24,25 @@
...
-	private String address = "*";
+	/**
+	 * Server should listen to any IPv4 and IPv6 address.
+	 */
+	public static final String ANY_IP_ADDRESS = "*";
+
+	/**
+	 * Server should listen to any IPv4 address.
+	 */
+	public static final String ANY_IPv4_ADDRESS = "0.0.0.0";
```
onobc, line 35: **[UNRELATED]** We are not using these last 2 yet. It helped me to see the constants though, named accordingly. I am happy to remove them until we actually use them though.

PR #24, opened 2024-10-06
Add keep-alive and inbound limits server properties

### LINKED ISSUE #6: Add more properties to GrpcServerProperties
Stuff that can be generically configured via a `ServerBuilder` (but excluding Spring Cloud features for now)

onobc:
**Commit 1: Add keep-alive and inbound limits server properties**

Resolves #6

> [!NOTE]
> I piggybacked the following feature on commit 2 (it is related and builds upon the first commit)

**Commit 2: Generate metadata for config props**
Adds the `spring-boot-configuration-processor` to enable automatic metadata
generation so our config props play well in IDEs.

Also updates the default values for duration properties to use
`Duration.of<Unit>(long)` instead of `Duration.of(long, DurationUnit)` as
Spring Boot config processor can not determine a default for the config
metadata in the latter format.

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/DefaultServerFactoryPropertyMapper.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/DefaultServerFactoryPropertyMapper.java
@@ -25,6 +25,7 @@ import io.grpc.ServerBuilder;
 
 import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.grpc.server.DefaultGrpcServerFactory;
+import org.springframework.util.unit.DataSize;
 
 /**
  * Helper class used to map {@link GrpcServerProperties} to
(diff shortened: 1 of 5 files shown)


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/DefaultServerFactoryPropertyMapper.java
```diff
@@ -46,10 +47,39 @@ class DefaultServerFactoryPropertyMapper<T extends ServerBuilder<T>> {
...
+		customizeKeepAlive(serverBuilder, mapper);
+		customizeInboundLimits(serverBuilder, mapper);
+	}
+
+	/**
+	 * Map the keep-alive properties to the server factory's server builder.
+	 * @param serverBuilder the builder
+	 * @param mapper the property mapper
+	 */
+	void customizeKeepAlive(T serverBuilder, PropertyMapper mapper) {
```
onobc, line 60: > [!NOTE]
> This property mapper abstraction insulates us from much of the duplication in the shaded/non-shaded concrete factory impls that we see in the grpc ecosystem version implemenation.

dsyer:
Seems like you might need to rebase on top of #23?

onobc:
> Seems like you might need to rebase on top of #23?

Thanks @dsyer - done.

PR #25, opened 2024-10-07
Generate config props docs

onobc:
> [!CAUTION]
> Review after #24 as this PR is built on top of the 2 commits in #24 

This adds automatic config props document generation.
Screenshot 2024-10-06 at 22 33 37

Screenshot 2024-10-06 at 22 33 45

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
@@ -1,0 +1,18 @@
+|===
+|Name | Default | Description
+
+|spring.grpc.client.channels |  | 
+|spring.grpc.server.address | `+++*+++` | Server address to bind to. The default is any IP address ('*').
+|spring.grpc.server.keep-alive.max-age |  | Maximum time a connection may exist before being gracefully terminated (default infinite).
+|spring.grpc.server.keep-alive.max-age-grace |  | Maximum time for graceful connection termination (default infinite).
+|spring.grpc.server.keep-alive.max-idle |  | Maximum time a connection can remain idle before being gracefully terminated (default infinite).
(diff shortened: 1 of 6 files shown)


spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
```diff
@@ -0,0 +1,18 @@
+|===
```
onobc, line 1: This file is automatically generated. Should we even check it in? I notice other spring-cloud repos are checking it in. I believe having it checked in avoids annoying red warnings coming from the consuming adocs that are referencing it.

spring-grpc-docs/src/main/java/org/springframework/grpc/internal/ConfigurationPropertiesAsciidocGenerator.java
```diff
@@ -0,0 +1,157 @@
...
+
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
+import org.springframework.util.StreamUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * Generate Asciidoc for configuration properties.
+ * <p>
+ * Copied from 'spring-cloud-build' to avoid direct dependency on Spring Cloud.
```
onobc, line 40: I did not want to put a direct dependency on `spring-cloud-build` just for this. I therefore copy/pasta shift/lift etc..

I will take a pass through this code in a subsequent PR as it can be simplified greatly and get a little 2024 makeover.

dsyer:
Might need to be rebased and force pushed your end?

onobc:
> Might need to be rebased and force pushed your end?

Should be good now @dsyer . Thanks.

Issue #29, opened 2024-10-18
Regularize server factories


dsyer:
The server factory has a different notion of what an "address" is to the client factory. This is awkward.

I don't like the way the `DefaultGrpcServerFactory` takes an address (which is actually a host) and a port in the constructor. It also seems like we should aim to use `Grpc.newServerBuilderForPort(port, creds)` instead  of the netty-specific versions. The netty-specific parts of the factory could then be isolated to creating the `ServerCredentials`, in principle. In practice I don't know if that would easily allow the "unix:" pseudo-scheme that we currently support.

If we get this right it should make it easier to add other server implementations (there's an okhttp one).

dsyer:
Fixed by #30

PR #30, opened 2024-10-18
Move some code around so factory gets address not port

dsyer:
See #29 for background

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryConfigurations.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryConfigurations.java
@@ -18,7 +18,7 @@ package org.springframework.grpc.autoconfigure.server;
 
 import java.util.List;
 
-import javax.net.ssl.SSLException;
+import javax.net.ssl.KeyManagerFactory;
 
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -28,17 +28,13 @@ import org.springframework.boot.ssl.SslBundle;
 import org.springframework.boot.ssl.SslBundles;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.grpc.server.DefaultGrpcServerFactory;
 import org.springframework.grpc.server.GrpcServerFactory;
 import org.springframework.grpc.server.NettyGrpcServerFactory;
 import org.springframework.grpc.server.ServerBuilderCustomizer;
 import org.springframework.grpc.server.ShadedNettyGrpcServerFactory;
 
 import io.grpc.BindableService;
-import io.grpc.ServerBuilder;
-import io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.NettyServerBuilder;
-import io.netty.handler.ssl.SslContextBuilder;
 
 /**
  * Configurations for {@link GrpcServerFactory gRPC server factories}.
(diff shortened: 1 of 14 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/util/GrpcUtils.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.grpc.util;
+
+public class GrpcUtils {
```
onobc, line 18: I think this is `public` mostly because it needs to be used by other packages in this project. 
I would rather not open this up to the world because as you know once its out there...
A freeing move may be to add the word `internal` into the package name which is a signal to the world that it is not a public api (yet). 

The fall-through approach to extract down to the numeric port is brilliant and pragmatic (very important in inception) and it solves our current needs. However, the moment we open it up for external use the edge-casey bugs will come flowing in. Long term I would imagine a regex impl that handles all the variations of bad user input. 

I don't want to figure out that magic regex now. 😄 

Wdyt?

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
```diff
@@ -65,19 +67,35 @@ public class GrpcServerProperties {
 
 	private final KeepAlive keepAlive = new KeepAlive();
 
+	private String address;
```
onobc, line 70: Do you mind adding a comment here - this will show up in the generated config props.

This `host and port OR address` reminds me of [MongoProperties](https://github.com/spring-projects/spring-boot/blob/37ae78ea73e043dbe8b605c586e3cc845a9d2240/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoProperties.java#L72). I like the idea of `deriveAddress` and leaving the properties alone `host,port,address`. I do ❤️ the validation that you can't set one and then the other (cake and eat it too) that you have done.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
```diff
@@ -35,13 +37,13 @@ public class GrpcServerProperties {
...
 	 */
-	private String address = ANY_IP_ADDRESS;
+	private String host = ANY_IP_ADDRESS;
 
 	/**
 	 * Server port to listen on. When the value is 0, a random available port is selected.
 	 * The default is 9090.
 	 */
-	private int port = 9090;
+	private int port = GrpcUtils.DEFAULT_PORT;
```
onobc, line 46: The generated config props is pretty smart, but it does not traverse class constants and leaves [default empty](https://github.com/spring-projects-experimental/spring-grpc/pull/30/files#diff-243856d43cde6a15728d5897a413d1f7c99b2743e8db88d938ea50889cd6ba0eR30). 

<sub>From [here](https://docs.spring.io/spring-boot/specification/configuration-metadata/annotation-processor.html)</sub>
> <sub>The annotation processor applies a number of heuristics to extract the default value from the source model. Default values have to be provided statically. In particular, do not refer to a constant defined in another class. Also, the annotation processor cannot auto-detect default values for Enums and Collections.</sub>

In order to get this default value into the metadata, we just need to add this to the [add. metadata](https://docs.spring.io/spring-boot/specification/configuration-metadata/annotation-processor.html#appendix.configuration-metadata.annotation-processor.adding-additional-metadata). Not sure if we have already created an add. metadata file. If you don't want to mess w/ this I am more than happy to handle this in a polish and merge. Just lmk

onobc (changes requested):
I am super excited about this huge simplification! I too sensed something w/ the port in `<init>` but could not put my finger on it. I have a few comments/suggestions, but other than that LGTM. Thanks @dsyer

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
```diff
@@ -18,9 +18,11 @@
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 
+import org.checkerframework.checker.units.qual.s;
```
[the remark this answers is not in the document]
dsyer: Weird. I don't see that. I'm going to ignore it.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
dsyer: I don't know how to do that yet. We'll do it later.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryConfigurations.java
```diff
@@ -99,56 +77,18 @@ static class NettyServerFactoryConfiguration {
...
-			else {
-				return builder -> {
-				};
+				keyManager = bundle.getManagers().getKeyManagerFactory();
 			}
-		}
-
-	}
-
-	@Configuration(proxyBeanMethods = false)
```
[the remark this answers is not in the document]
dsyer: I think we can put it back if we ever need it. I suppose it probably would support okhttp even now (but I think only if there is an SSL bundle), but I'd want a sample to support that assumption.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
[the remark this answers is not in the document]
dsyer: Indeed

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerPropertiesTests.java
```diff
@@ -48,11 +48,11 @@ class BaseProperties {
 		@Test
 		void bind() {
 			Map<String, String> map = new HashMap<>();
```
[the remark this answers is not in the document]
dsyer: Done

spring-grpc-core/src/main/java/org/springframework/grpc/util/GrpcUtils.java
dsyer: We can change the package name.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
onobc: I'll drop one in - no worries.

PR #33, opened 2024-10-20
Add additional config props metadata file

onobc:
This commit adds the additional-spring-configuration-metadata.json file to the autoconfigure module so that we can add default values for properties whose defaults can not be determined.


onobc:
@dsyer This is the thing I mentioned [here](https://github.com/spring-projects-experimental/spring-grpc/pull/30#discussion_r1807471550). Going to merge as its trivial.

PR #36, opened 2024-10-21
Add spring-grpc-spring-boot-starter module

onobc:
This PR does the following:

* Adds a spring-grpc-spring-boot-starter module as described [here](https://docs.spring.io/spring-boot/reference/features/developing-auto-configuration.html#features.developing-auto-configuration.custom-starter.starter-module).
* Adjusts the scope of the dependencies in the spring-grpc-autoconfigure module as [recommended here](https://docs.spring.io/spring-boot/reference/features/developing-auto-configuration.html#features.developing-auto-configuration.custom-starter.autoconfigure-module). 

At first glance, the autoconfigure is simple enough to just combine the starter + autoconfigure in a single module. However, I can envision us having multiple starters in the future (netty, shaded netty, okio, etc...). 

As it currently stands, our default opinion is "netty" over "shaded netty". 

1. Should that be our default?
2. Should we go ahead and name this default starter w/ according suffix, something like the following:?

- spring-grpc-spring-boot-starter-shaded-netty
- spring-grpc-spring-boot-starter-netty
- spring-grpc-spring-boot-starter-okio

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To create a simple gRPC server, you can use the Spring Boot starter. For Maven:
 ```xml
 <dependency>
 	<groupId>org.springframework.grpc</groupId>
-	<artifactId>spring-grpc-spring-boot-autoconfigure</artifactId>
+	<artifactId>spring-grpc-spring-boot-starter</artifactId>
 	<version>0.1.0-SNAPSHOT</version>
 </dependency>
 ```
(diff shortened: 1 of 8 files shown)


dsyer:
This change was coming sooner or later, so I suppose it might as well be now.

Personally I like to have a default (no need to engage brain). But I prefer netty to shaded netty just because of awkwardness with IDEs not showing source code - also we'd like to get it working in a webflux app so we'd want the same netty as reactor-netty for that. So I can see why explicitly naming all the starters (and no default) makes sense, but I just can't make myself like it.

onobc:
I am good w/ the no suffix as the default as well. This is what spring-cloud-gateway and others do as well. We can append suffixes as needed when the time comes. Thanks for the review and merge.

PR #37, opened 2024-10-23
Add autoconfiguration for gRPC servlet

### LINKED ISSUE #18: Support using the same port for Spring MVC and gRPC
In some environments, such as PaaS or Knative, using multiple ports may not be the default.
It would be very convenient to be able to serve plain Spring MVC and gRCP on 8080.

I don't know what it uses internally, but Armeria has made it possible.
https://armeria.dev/docs/advanced-spring-boot-integration

dsyer:
In a web application, if there are BindableServices we use
a servlet instead of a netty-based server.

Fixes #18

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
@@ -12,23 +12,23 @@
 |spring.grpc.client.default-channel.max-inbound-message-size |  | 
 |spring.grpc.client.default-channel.max-inbound-metadata-size |  | 
 |spring.grpc.client.default-channel.negotiation-type |  | The negotiation type for the channel. Default is {@link NegotiationType#PLAINTEXT}.
-|spring.grpc.client.default-channel.secure | `+++true+++` | Flag to say that strict SSL checks are not enabled (so the remote certificate could be anonymous).
+|spring.grpc.client.default-channel.secure |  | Flag to say that strict SSL checks are not enabled (so the remote certificate could be anonymous).
 |spring.grpc.client.default-channel.ssl.bundle |  | SSL bundle name.
 |spring.grpc.client.default-channel.ssl.enabled |  | Whether to enable SSL support. Enabled automatically if "bundle" is provided unless specified otherwise.
 |spring.grpc.client.default-channel.user-agent |  | 
 |spring.grpc.server.address |  | The address to bind to. could be a host:port combination or a pseudo URL like static://host:port. Can not be set if host or port are set independently.
-|spring.grpc.server.host | `+++*+++` | Server address to bind to. The default is any IP address ('*').
+|spring.grpc.server.host |  | Server address to bind to. The default is any IP address ('*').
 |spring.grpc.server.keep-alive.max-age |  | Maximum time a connection may exist before being gracefully terminated (default infinite).
 |spring.grpc.server.keep-alive.max-age-grace |  | Maximum time for graceful connection termination (default infinite).
 |spring.grpc.server.keep-alive.max-idle |  | Maximum time a connection can remain idle before being gracefully terminated (default infinite).
-|spring.grpc.server.keep-alive.permit-time | `+++5m+++` | Maximum keep-alive time clients are permitted to configure (default 5m).
-|spring.grpc.server.keep-alive.permit-without-calls | `+++false+++` | Whether clients are permitted to send keep alive pings when there are no outstanding RPCs on the connection (default false).
-|spring.grpc.server.keep-alive.time | `+++2h+++` | Duration without read activity before sending a keep alive ping (default 2h).
-|spring.grpc.server.keep-alive.timeout | `+++20s+++` | Maximum time to wait for read activity after sending a keep alive ping. If sender does not receive an acknowledgment within this time, it will close the connection (default 20s).
-|spring.grpc.server.max-inbound-message-size | `+++4194304B+++` | Maximum message size allowed to be received by the server (default 4MiB).
-|spring.grpc.server.max-inbound-metadata-size | `+++8192B+++` | Maximum metadata size allowed to be received by the server (default 8KiB).
-|spring.grpc.server.port | `+++9090+++` | Server port to listen on. When the value is 0, a random available port is selected. The default is 9090.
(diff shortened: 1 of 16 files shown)


spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
```diff
@@ -12,23 +12,23 @@
...
 |spring.grpc.client.default-channel.max-inbound-metadata-size |  | 
 |spring.grpc.client.default-channel.negotiation-type |  | The negotiation type for the channel. Default is {@link NegotiationType#PLAINTEXT}.
-|spring.grpc.client.default-channel.secure | `+++true+++` | Flag to say that strict SSL checks are not enabled (so the remote certificate could be anonymous).
+|spring.grpc.client.default-channel.secure |  | Flag to say that strict SSL checks are not enabled (so the remote certificate could be anonymous).
 |spring.grpc.client.default-channel.ssl.bundle |  | SSL bundle name.
 |spring.grpc.client.default-channel.ssl.enabled |  | Whether to enable SSL support. Enabled automatically if "bundle" is provided unless specified otherwise.
 |spring.grpc.client.default-channel.user-agent |  | 
 |spring.grpc.server.address |  | The address to bind to. could be a host:port combination or a pseudo URL like static://host:port. Can not be set if host or port are set independently.
-|spring.grpc.server.host | `+++*+++` | Server address to bind to. The default is any IP address ('*').
+|spring.grpc.server.host |  | Server address to bind to. The default is any IP address ('*').
```
onobc, line 20: Do you mind re-running `./mvnw clean install` locally to update these config props? I think they must have been run in some inconsistent state or something to produce this w/ no defaults. I ran locally and it re-populated.

spring-grpc-spring-boot-autoconfigure/pom.xml
```diff
@@ -45,6 +45,21 @@
...
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.tomcat.embed</groupId>
+			<artifactId>tomcat-embed-core</artifactId>
+			<optional>true</optional>
+		</dependency>
```
onobc, line 57: There is nothing special about Tomcat and the GrpcServlet (i.e. this would work w/ Jetty or Undertow as well) correct?

If so, we can make this server agnostic by replacing this w/ 

```suggestion
    <dependency>
      <groupId>jakarta.servlet</groupId>
      <artifactId>jakarta.servlet-api</artifactId>
      <optional>true</optional>
    </dependency>
```
Wdyt?

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryAutoConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+	@ConditionalOnNotWebApplication
+	@Import({ GrpcServerFactoryConfigurations.ShadedNettyServerFactoryConfiguration.class,
+			GrpcServerFactoryConfigurations.NettyServerFactoryConfiguration.class })
+	static class GrpcServerFactoryConfiguration {
+
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnWebApplication
+	static class GrpcServletConfiguration {
```
onobc, line 60: There are limited options for the `ServletServerBuilder` so providing a custom server factory is overkill since the `ServletAdapter#init `does the same thing we would be doing:

```java
  public GrpcServlet(List<? extends BindableService> bindableServices) {
    this(loadServices(bindableServices));
  }

  private static ServletAdapter loadServices(List<? extends BindableService> bindableServices) {
    ServletServerBuilder serverBuilder = new ServletServerBuilder();
    bindableServices.forEach(serverBuilder::addService);
    return serverBuilder.buildServletAdapter();
  }
```

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryAutoConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+ * @author David Syer
+ * @author Chris Bono
+ */
+@AutoConfiguration
+@AutoConfigureOrder(Ordered.HIGHEST_PRECEDENCE)
+@ConditionalOnClass(BindableService.class)
+public class GrpcServerFactoryAutoConfiguration {
+
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnNotWebApplication
```
onobc, line 51: <sub><sub>**Legend:** `Tomcat/Jetty/Undertow` --> `T/J/UT`</sub></sub>
--------------

Before this change, we could run Spring gRPC in a servlet web app (T/J/UT) or Reactive web app (Netty). <sub>If its a Reactive web app it has to be Netty of course.</sub>

With this change, we can not run Spring gRPC in any webapp.

Should we only enforce a limitation when the user desires the servlet to be available? And even in that case, if the `server.port` and `spring.grpc.server.port` are different we should be able to run Spring gRPC and GrpcServlet in a servlet web app (T/J/UT) or in a reactive web app (Netty).

If we want the ports to be the same then the Spring gRPC server must be disabled which goes against the goal of the motivating issue #18 , serving both MVC and gRPC from same port (e.g. `9090`).

onobc (changes requested):
This is cool stuff @dsyer . I have added a few comments/concerns.

spring-grpc-spring-boot-autoconfigure/pom.xml
dsyer: It doesn't matter much, since it's optional, but I can make that change.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryAutoConfiguration.java
dsyer: > With this change, we can not run Spring gRPC in any webapp.

I don't think that's a very common scenario. I kind of expect that we will eventually support it through an `*.enabled` flag. I didn't want to do that until someone asks for it because it will make the `@Conditionals` a lot more complicated.

Issue #4, opened 2024-09-11
Automatically pick up global ServiceInterceptors


dsyer:
Maybe with an annotation (is it necessary?).

PR #41, opened 2024-10-27
#1: Autoconfigure metric registry support

### LINKED ISSUE #1: Support for Micrometer observations collection
E.g. add `GrpcServerMetricAutoConfiguration`.

hmtang:
Signed-off-by: Sunny <hmtang@gmail.com>
[resolves #1]

### DIFF
--- a/spring-grpc-core/pom.xml
+++ b/spring-grpc-core/pom.xml
@@ -68,11 +68,21 @@
 		</dependency>
 
 		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-registry-prometheus</artifactId>
+		</dependency>
+
+		<dependency>
 			<groupId>org.junit.jupiter</groupId>
 			<artifactId>junit-jupiter</artifactId>
 			<version>${junit.version}</version>
 			<scope>test</scope>
 		</dependency>
+
 
 	</dependencies>
 
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
@@ -23,16 +23,11 @@ import java.util.Set;
 
 import com.google.common.collect.Lists;
 
-import io.grpc.Grpc;
-import io.grpc.InsecureServerCredentials;
-import io.grpc.Server;
-import io.grpc.ServerBuilder;
-import io.grpc.ServerCredentials;
-import io.grpc.ServerProvider;
-import io.grpc.ServerServiceDefinition;
+import io.grpc.*;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.grpc.internal.GrpcUtils;
 
 /**
@@ -55,6 +50,9 @@ public class DefaultGrpcServerFactory<T extends ServerBuilder<T>> implements Grp
 	private final String address;
 
 	private final List<ServerBuilderCustomizer<T>> serverBuilderCustomizers;
+
+	@Autowired
+	private List<ServerInterceptor> serverInterceptors;
 
 	public DefaultGrpcServerFactory(String address, List<ServerBuilderCustomizer<T>> serverBuilderCustomizers) {
 		this.address = address;
@@ -83,7 +81,11 @@ public class DefaultGrpcServerFactory<T extends ServerBuilder<T>> implements Grp
 	 */
 	@SuppressWarnings("unchecked")
 	protected T newServerBuilder() {
-		return (T) Grpc.newServerBuilderForPort(port(), credentials());
+		var builder = Grpc.newServerBuilderForPort(port(), credentials());
+		for (ServerInterceptor interceptor : serverInterceptors) {
+			builder.intercept(interceptor);
+		}
+		return (T) builder;
 	}
 
 	/**
--- a/spring-grpc-dependencies/pom.xml
+++ b/spring-grpc-dependencies/pom.xml
@@ -54,6 +54,7 @@
 		<grpc.version>1.63.2</grpc.version>
 		<protobuf-java.version>3.25.5</protobuf-java.version>
 		<google-common-protos.version>2.46.0</google-common-protos.version>
+		<micrometer.version>1.13.6</micrometer.version>
 	</properties>
 
 	<dependencyManagement>
@@ -111,6 +112,16 @@
 				<groupId>com.google.api.grpc</groupId>
 				<artifactId>proto-google-common-protos</artifactId>
 				<version>${google-common-protos.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.micrometer</groupId>
+				<artifactId>micrometer-core</artifactId>
+				<version>${micrometer.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.micrometer</groupId>
+				<artifactId>micrometer-registry-prometheus</artifactId>
+				<version>${micrometer.version}</version>
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
@@ -15,6 +15,10 @@
  */
 package org.springframework.grpc.autoconfigure.server;
 
+import io.grpc.ServerInterceptor;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.grpc.MetricCollectingServerInterceptor;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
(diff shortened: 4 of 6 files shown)


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfiguration.java
```diff
@@ -69,4 +73,15 @@ ServerBuilderCustomizers serverBuilderCustomizers(ObjectProvider<ServerBuilderCu
 		return new ServerBuilderCustomizers(customizers.orderedStream().toList());
 	}
 
+	@ConditionalOnMissingBean
```
onobc, line 76: We do not want to impose micrometer on user's. Only if they decide to pull Micrometer onto their classpath should we react to that. Let's remove this `MeterRegistry` bean.

Also, we will want to split this off into its own auto-configuration (e.g. `GrpcServerMetricAutoConfiguration`) and then guard its loading based on whether user has Micrometer on the classpath. The gRPC ecocystem h as a great example [here](https://github.com/yidongnan/grpc-spring-boot-starter/blob/242ee88dc8bc651479a8a8bb7c03ff0ec65aacbb/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcServerMetricAutoConfiguration.java#L57-L77).

spring-grpc-core/pom.xml
```diff
@@ -67,13 +67,23 @@
 			<artifactId>grpc-stub</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-core</artifactId>
+		</dependency>
+		<dependency>
```
onobc, line 74: We do not have an opinion on using Prometheus, let's remove this dependency.

spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
```diff
@@ -56,6 +51,9 @@ public class DefaultGrpcServerFactory<T extends ServerBuilder<T>> implements Grp
 
 	private final List<ServerBuilderCustomizer<T>> serverBuilderCustomizers;
 
+	@Autowired
```
onobc, line 54: Since we do not yet have support for interceptors, lets not add that in this PR as I am actively working on #4. Instead you can use a `ServerBuilderCustomizer` as follows:

```java
@Bean
<T extends ServerBuilder<T>> ServerBuilderCustomizer<T> metricsInterceptor() {
	return (serverBuilder) -> serverBuilder.intercept(new MyXyInterceptor()));
}
```

If you add the above customizer in the auto-configuration then it will get applied at server build time.

spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
```diff
@@ -23,16 +23,11 @@
...
 import com.google.common.collect.Lists;
 
-import io.grpc.Grpc;
-import io.grpc.InsecureServerCredentials;
-import io.grpc.Server;
-import io.grpc.ServerBuilder;
-import io.grpc.ServerCredentials;
-import io.grpc.ServerProvider;
-import io.grpc.ServerServiceDefinition;
+import io.grpc.*;
```
onobc, line 26: Please remove the wildcard imports. We need to update our formatter to balk about this.

spring-grpc-dependencies/pom.xml
```diff
@@ -112,6 +113,16 @@
 				<artifactId>proto-google-common-protos</artifactId>
 				<version>${google-common-protos.version}</version>
 			</dependency>
+			<dependency>
```
onobc, line 116: Please replace this w/ `micrometer-bom` which handles all `io.micrometer` version constraints.

spring-grpc-dependencies/pom.xml
```diff
@@ -112,6 +113,16 @@
...
+				<groupId>io.micrometer</groupId>
+				<artifactId>micrometer-core</artifactId>
+				<version>${micrometer.version}</version>
+			</dependency>
+			<dependency>
```
onobc, line 121: Please remove prometheus, we have no opinion on using this.

onobc (changes requested):
Thanks @hmtang for the contribution. I have made a few request for changes. Let us know if you have any questions or need further clarifications - we are excited to have you on the project. Thank you.

hmtang:
Thanks a lot for the feedbacks. I would be out for a short trip in next week. I would look into this when I get back. Appreciate your time on the reviews.

onobc:
>I don't think we need MetricCollectingServerInterceptor if we support ObservationGrpcServerInterceptor which can output similar metrics and traces at the same time.

I agree w/ this @making . I think we should just add auto-configuration for the Micrometer Observations API (which gives metrics and tracing) and if users want to only use Micrometer metrics they can add it in via interceptor themselves. 

We have started taking this approach in our newer Spring projects (e.g. Spring Pulsar only provides support for Observations API). Most Spring projects were incepted prior to the availability of Observations API therefore Micrometer support was already in place. 

@dsyer do you agree w/ this?

dsyer:
100%

spring-grpc-core/pom.xml
```diff
@@ -67,13 +67,19 @@
 			<artifactId>grpc-stub</artifactId>
 		</dependency>
 
+		<dependency>
```
dsyer, line 70: This should be optional? Or maybe it's already present transitively from Spring (so we don't need to declare it at all)?

hmtang: io.micrometer.core.instrument.binder.grpc.ObservationGrpcServerInterceptor would be needed for the AutoConfiguration and that's not included I guess as I found this is breaking if I take it out.

onobc (changes requested):
Nice work @hmtang ! A have added a few comments/suggestions. 

Besides the suggestions, once we get the base configuration in this PR we will definitely want to document how to enable observations in a Spring gRPC application. There are quite a few moving pieces and it can be confusing. It does not have to be in this PR though as it may get large.

We did this exact thing for [Spring Pulsar docs](https://docs.spring.io/spring-pulsar/reference/reference/observability.html#_auto_configuration_with_spring_boot) and we can clone/own those or just point to the SB docs on this topic.

We may want to create a sample that uses Micrometer observations as well (we did the same thing in Spring Pulsar [here](https://github.com/spring-projects/spring-pulsar/blob/62fff05db1dc4b3c467e04c6c1c4f96540782389/spring-pulsar-sample-apps/sample-failover-custom-router/build.gradle#L23-L26).

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerMetricAutoConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.context.annotation.Bean;
+import org.springframework.grpc.server.ServerBuilderCustomizer;
+
+@AutoConfiguration
+@ConditionalOnClass(ObservationGrpcServerInterceptor.class)
+@ConditionalOnBean(ObservationRegistry.class)
+public class GrpcServerMetricAutoConfiguration {
```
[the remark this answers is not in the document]
onobc: A follow up comment about the name of the above property... 

I think we should name this property `spring.grpc.server.observation.enabled` as it is not really the same thing as the actuator `management.<system>.tracing` property stem. This is really more of a flag as to whether we should instrument the gRPC server or not. @dsyer wdyt?

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerMetricAutoConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+
+	@Bean
+	public ServerInterceptor observationGrpcServerInterceptor(ObservationRegistry observationRegistry) {
+		return new ObservationGrpcServerInterceptor(observationRegistry);
+	}
+
+	@Bean
+	<T extends ServerBuilder<T>> ServerBuilderCustomizer<T> metricsInterceptor(
+			ServerInterceptor observationGrpcServerInterceptor) {
+		return (serverBuilder) -> serverBuilder.intercept(observationGrpcServerInterceptor);
```
making, line 26: This definition has broader effects than the purpose of this PR, since it adds arbitrary `ServerInterceptor`s to the `ServerBuilder`, which should be defined in a separate class.

onobc: Hi @making , This is a temporary means to apply the interceptor to the builder via auto-configuration. 
This will be replaced by marking the interceptor bean above w/ `@GrpcGlobalServerInterceptor`. I am working on that now.

making: Thanks for the clarification!

hmtang:
Thanks for the reviews. I have pushed a new changes for the requested changes.

dsyer:
"This branch cannot be rebased safely" according to github, so you need to rebase. Probably a good idea to squash your commits into one as well and force push back to your fork.

onobc:
Thanks again for the great contribution @hmtang . I am closing this PR in favor of https://github.com/spring-projects-experimental/spring-grpc/commit/7fed50a42008713703c9004da12b6880eb6223c3.

I added a follow-on commit to your commit (see details in commit message if you are interested. I add a couple more tests).

onobc:
> "This branch cannot be rebased safely" according to github, so you need to rebase. Probably a good idea to squash your commits into one as well and force push back to your fork.

Sorry @dsyer , 
I went ahead and handled that locally when I added polish commit. Next time I will assign the PR to myself to give a signal.

PR #42, opened 2024-10-28
Support `ServerBuilderCustomizer` and `spring.grpc.server.max-inbound-message-size` property for `GrpcServlet`

making:
Currently `GrpcServlet` is not customizable (e.g. Adding interceptors) 
This PR supports `ServerBuilderCustomizer` and `spring.grpc.server.max-inbound-message-size` property for `GrpcServlet`.

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
@@ -16,13 +16,9 @@
 
 package org.springframework.grpc.autoconfigure.server;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import io.grpc.BindableService;
 import io.grpc.ServerServiceDefinition;
 import io.grpc.internal.GrpcUtil;
-import io.grpc.servlet.jakarta.GrpcServlet;
-import io.grpc.servlet.jakarta.ServletAdapter;
 import io.grpc.servlet.jakarta.ServletServerBuilder;
 import org.junit.jupiter.api.Test;
 
@@ -35,7 +31,9 @@ import org.springframework.grpc.server.ServerBuilderCustomizer;
 import org.springframework.util.unit.DataSize;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 /**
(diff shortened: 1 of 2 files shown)


spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
```diff
@@ -90,7 +74,40 @@ void whenNoBindableServicesRegisteredAutoConfigurationIsSkipped() {
 
 	@Test
 	void whenWebApplicationServletIsAutoConfigured() {
-		this.contextRunner().run((context) -> assertThat(context).getBean(ServletRegistrationBean.class).isNotNull());
+		this.contextRunner().run((context) -> {
```
onobc, line 77: I think we can leave this test the way that it was and verify default property values below.

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
```diff
@@ -90,7 +74,40 @@ void whenNoBindableServicesRegisteredAutoConfigurationIsSkipped() {
...
+				.isInstanceOf(GrpcServlet.class)
+				.extracting("servletAdapter")
+				.isInstanceOf(ServletAdapter.class)
+				.extracting("maxInboundMessageSize")
+				.isEqualTo(GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE);
+		});
+	}
+
+	@Test
+	void whenCustomizerIsRegistered() {
```
onobc, line 90: I think we can simplify this just a bit w/ a mock - something like:
```java
	@Test
	void whenCustomizerIsRegistered() {
		ServerBuilderCustomizer<ServletServerBuilder> customizer = mock();
		this.contextRunner()
			.withBean(ServerBuilderCustomizer.class, () -> customizer)
			.run(context -> verify(customizer).customize(any(ServletServerBuilder.class)));
	}
```

Wdyt?

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
```diff
@@ -90,7 +74,40 @@ void whenNoBindableServicesRegisteredAutoConfigurationIsSkipped() {
...
+		AtomicBoolean invoked = new AtomicBoolean(false);
+		ServerBuilderCustomizer<ServletServerBuilder> customizer = serverBuilder -> invoked.set(true);
+		this.contextRunner().withBean(ServerBuilderCustomizer.class, () -> customizer).run(context -> {
+			assertThat(context).getBean(ServletRegistrationBean.class).isNotNull();
+			assertThat(invoked.get()).isTrue();
+		});
+	}
+
+	@Test
+	void whenMaxInboundMessageSizeIsConfigured() {
```
onobc, line 100: As mentioned above, I think adding another variant here that tests the property value when set and when not set would be good. Something like:

```java
	@Test
	void whenMaxInboundMessageSizeIsSetThenItIsUsed() {
		this.contextRunner()
			.withPropertyValues("spring.grpc.server.max-inbound-message-size=10KB")
			.run(context -> assertThat(context).getBean(ServletRegistrationBean.class)
				.hasFieldOrPropertyWithValue("servlet.servletAdapter.maxInboundMessageSize",
						Math.toIntExact(DataSize.ofKilobytes(10).toBytes())));
	}

	@Test
	void whenMaxInboundMessageSizeIsNotSetThenDefaultIsUsed() {
		this.contextRunner()
			.run((context) -> assertThat(context).getBean(ServletRegistrationBean.class)
				.hasFieldOrPropertyWithValue("servlet.servletAdapter.maxInboundMessageSize",
						GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE));
	}
```
Also, notice the use of `hasFieldOrPropertyWithValue` to avoid the nested drill-down to get to the property value. 

Wdyt?

onobc (changes requested):
Nice work @making . This servlet factory is the only one that we did not create its own `GrpcServerFactory` impl so this is fine to add the property mapping and customizers here for now. If we later create its own server factory we can move this logic then (e.g. [like here](https://github.com/spring-projects-experimental/spring-grpc/blob/5ab408b62658bb0707ae5f29099dcf7f72b8ae4d/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryConfigurations.java#L99)).

A few minor suggestions / tips on the included tests but once we go through those this will be ready to merge. 

BTW - thank you for adding the excellent test coverage.

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
making: make sense. fixed in 2f50e5c27732722dd6027127135d1673f1503a59

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
making: make sense. fixed in 2f50e5c27732722dd6027127135d1673f1503a59

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServletAutoConfigurationTests.java
making: make sense. fixed in 2f50e5c27732722dd6027127135d1673f1503a59

making:
> If we later create its own server factory we can move this logic then (e.g. [like here](https://github.com/spring-projects-experimental/spring-grpc/blob/5ab408b62658bb0707ae5f29099dcf7f72b8ae4d/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryConfigurations.java#L99)).

I also initially tried to create a ServletServerFactoryPropertyMapper, but since technically there is no "ServerFactory" for GrpcServlet and there were only a few configurable items, I decided to do the mapping inline now.

onobc:
Closing in favor of https://github.com/spring-projects-experimental/spring-grpc/commit/7da1e80b55825c14ef6e4fd6a502f6643b214f34

Issue #26, opened 2024-10-08
Add autoconfig for reflection service and friends


dsyer:
Probably all gated behind a boolean `enabled` flag. The configuration is easy, it's just deciding what to do by default, e.g. in the existing sample

```java
@Bean
public BindableService serverReflection() {
	return ProtoReflectionService.newInstance();
}
```

CyberZujo:
I'd like to help here. Can you assign it please. Thank you!

CyberZujo:
https://github.com/spring-projects-experimental/spring-grpc/pull/45
@onobc Can you please have a look and provide a feedback if I'm on the right track.
Thank you. 
Kind regards, Haris.

onobc:
Hi @CyberZujo , I will take a look sometime today. Thank you.

CyberZujo:
Everything good with this one?
@dsyer

nikita-rogovyi:
If while using `grpcurl` you see `server does not support the reflection API` but logs show `grpc.reflection.v1.ServerReflection` is registered, make sure that you use grpcurl version ≥1.9.0 — older versions only support the v1alpha API.

PR #45, opened 2024-11-03
Add autoconfig for reflection service and friends #26

CyberZujo:
**… - Add autoconfig for reflection service and friends #26**

- Created new Autoconfiguration for BindableService bean instantiation
- Bean creation triggered based on the presence of configuration property
- Config property: grpc.server.reflection.enabled=true/false

### DIFF
--- a/samples/grpc-server/src/main/resources/application.properties
+++ b/samples/grpc-server/src/main/resources/application.properties
@@ -1,2 +1,2 @@
 spring.application.name=grpc-server
-grpc.server.reflection.enabled=false
+spring.grpc.server.reflection.enabled=true
--- a/samples/grpc-server/src/main/resources/application.properties
+++ b/samples/grpc-server/src/main/resources/application.properties
@@ -1,1 +1,2 @@
 spring.application.name=grpc-server
+grpc.server.reflection.enabled=false
--- a/spring-grpc-spring-boot-autoconfigure/pom.xml
+++ b/spring-grpc-spring-boot-autoconfigure/pom.xml
@@ -88,7 +88,11 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-services</artifactId>
+        </dependency>
 
-	</dependencies>
+    </dependencies>
 
 </project>
--- a/spring-grpc-spring-boot-autoconfigure/pom.xml
+++ b/spring-grpc-spring-boot-autoconfigure/pom.xml
@@ -62,6 +62,11 @@
 		</dependency>
 		<dependency>
 			<groupId>io.grpc</groupId>
+			<artifactId>grpc-services</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>io.grpc</groupId>
 			<artifactId>grpc-netty-shaded</artifactId>
 			<optional>true</optional>
 		</dependency>
@@ -88,10 +93,6 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
-        <dependency>
-            <groupId>io.grpc</groupId>
-            <artifactId>grpc-services</artifactId>
-        </dependency>
 
     </dependencies>
 
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
@@ -27,4 +27,5 @@ public class GrpcServerReflectionAutoConfiguration {
 	public BindableService serverReflection() {
 		return ProtoReflectionService.newInstance();
 	}
+
 }
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
@@ -12,7 +12,8 @@ import io.grpc.protobuf.services.ProtoReflectionService;
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for gRPC Reflection service
  * <p>
- * spring.grpc.reflection.enabled=true must be present in configuration in order for the
+ * This auto-configuration is enabled by default. To disable it, set the configuration
+ * flag {spring.grpc.server.reflection.enabled=false} in your application properties.
  * auto-configuration to execute
  *
  * @author Haris Zujo
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
@@ -18,7 +18,7 @@ import io.grpc.protobuf.services.ProtoReflectionService;
  * @author Haris Zujo
  */
 @Configuration
-@ConditionalOnClass(BindableService.class)
+@ConditionalOnClass(ProtoReflectionService.class)
 public class GrpcServerReflectionAutoConfiguration {
 
 	@Bean
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
@@ -1,0 +1,30 @@
+package org.springframework.grpc.autoconfigure.server;
+
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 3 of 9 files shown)


samples/grpc-server/src/main/resources/application.properties
```diff
@@ -1 +1,2 @@
 spring.application.name=grpc-server
+grpc.server.reflection.enabled=false
```
dsyer, line 2: I think we actually need this to be true (and then we can remove the explicit `@Bean` definition in the app) so that the instructions in the README actually work as described.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for gRPC Reflection service
+ * <p>
+ * grpc.reflection.enabled=true must be present in configuration in order for the
+ * auto-configuration to execute
+ *
+ * @author Haris Zujo
+ */
+@Configuration
+@ConditionalOnClass(BindableService.class)
```
onobc, line 21: We will want to guard this a little bit wider as the `ProtoReflectionService` is optional. We can replace w/:
```suggestion
@ConditionalOnClass(ProtoReflectionService.class)
```

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+public class GrpcServerReflectionAutoConfiguration {
+
+	@Bean
+	@ConditionalOnProperty(name = "grpc.server.reflection.enabled", havingValue = "true")
```
onobc, line 25: We can move this conditional up to the class level as well. I would also add `matchIfMissing = true`.

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfigurationTests.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class GrpcServerReflectionAutoConfigurationTests {
+
+	private ApplicationContextRunner contextRunner() {
+		return new ApplicationContextRunner()
+			.withConfiguration(AutoConfigurations.of(GrpcServerReflectionAutoConfiguration.class))
+			.withBean("noopServerLifecylcle", GrpcServerLifecycle.class, Mockito::mock);
+	}
+
+	@Test
```
onobc, line 20: For the "binary" flag properties I like to add 3 test cases...
1. property not set at all - what does the default do
2. property set to true
3. property set to false.

Wdyt?

spring-grpc-spring-boot-autoconfigure/pom.xml
```diff
@@ -88,7 +88,11 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
```
onobc, line 91: I believe this needs to move up out of the "test" dependencies area and mark it w/ `<optional>true</optional>` which will result in it being usable in tests as well.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import io.grpc.BindableService;
+import io.grpc.protobuf.services.ProtoReflectionService;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for gRPC Reflection service
+ * <p>
+ * grpc.reflection.enabled=true must be present in configuration in order for the
```
onobc, line 15: Let's prefix this property w/ `spring.` and also please add an entry to the [additional metadata file](https://github.com/spring-projects-experimental/spring-grpc/blob/9cf02273548b9642087ae0d5ee0d4503ae7f054d/spring-grpc-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L5) that way auto-complete in editor will work for this property.

<sub>Background [here](https://docs.spring.io/spring-boot/specification/configuration-metadata/annotation-processor.html#appendix.configuration-metadata.annotation-processor.adding-additional-metadata)</sub>

onobc (changes requested):
Thanks for the contribution @CyberZujo . I have a added a few comments/suggestions. It is close to ready though.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ * auto-configuration to execute
+ *
+ * @author Haris Zujo
+ */
+@Configuration
```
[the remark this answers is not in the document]
dsyer: Might be better to but `@AutoConfigureBefore` on this one?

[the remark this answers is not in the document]
vuhp: Yes, I agree!

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
CyberZujo: Updated.

samples/grpc-server/src/main/resources/application.properties
CyberZujo: Updated.

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfigurationTests.java
CyberZujo: Done.

spring-grpc-spring-boot-autoconfigure/pom.xml
CyberZujo: Done.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
CyberZujo: Done.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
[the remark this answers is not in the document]
CyberZujo: Updated as well.

CyberZujo:
@onobc Updated the PR, please check! Thanks!

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
```diff
@@ -1,30 +1,29 @@
...
-import org.springframework.context.annotation.Configuration;
 
 import io.grpc.BindableService;
 import io.grpc.protobuf.services.ProtoReflectionService;
 
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for gRPC Reflection service
  * <p>
- * grpc.reflection.enabled=true must be present in configuration in order for the
+ * spring.grpc.reflection.enabled=true must be present in configuration in order for the
```
dsyer, line 15: This is no longer quite true right? It’s on by default (and hence not needed explicitly in the samples)?

CyberZujo: True, will update soon!

CyberZujo: * Auto-configuration enabled by default
 Should I replace with this kind of comment or?

dsyer: Makes sense. You can still mention the flag, but use “false” ad the example because that’s likely what most people will need.

CyberZujo: /**
 * {@link EnableAutoConfiguration Auto-configuration} for gRPC Reflection service
 * <p>
 * This auto-configuration is enabled by default. To disable it, set the configuration
 * flag {spring.grpc.server.reflection.enabled=false} in your application properties.
 * auto-configuration to execute
 *
 * @author Haris Zujo
 */
 
 
 Updated

dsyer:
Merged in d61af1a

Issue #5, opened 2024-09-11
Add specific interceptors (not global) to a gRPC service


dsyer:
Maybe an annotation? Might need to refer to bean instances (c.f. interceptors and listeners in `@Retryable` or Spring Batch).

PR #46, opened 2024-11-06
Add GlobalServerInterceptor to service discovery

### LINKED ISSUE #4: Automatically pick up global ServiceInterceptors

onobc:
This commit builds upon the previously added service discoverer by adding support for finding all server interceptor beans that are marked w/ the newly added `@GlobalServerInterceptor` annotation.

Resolves #4

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/pages/server.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/pages/server.adoc
@@ -86,3 +86,23 @@ dependencies {
 The `spring.grpc.server.*` properties will be ignored in facour of the regular `server.*` properties in this case.
 The servlet that is created is mapped to process HTTP POST requests to the paths defined by the registered services, as `/<service-name>/*`.
 Clients can connect to the server using that path, which is what any gRPC client library will do.
+
+[[server-interceptor]]
+== Server Interceptors
+
+=== Global
+To add a server interceptor to be applied to all services you can simply register a server interceptor bean and then annotate it with `@GlobalServerInterceptor`.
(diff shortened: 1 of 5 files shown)


dsyer:
I'm confused. Maybe this is just a stepping stone? But the docs don't seem to make sense for the non-global case - `ServerBuilder.intercept()` is a method to add a global interceptor and you use it in the example of a non-global, unless I'm reading it wrong somehow. I think you need to create and wrap a service definition with `ServerInterceptors`.

onobc:
> I'm confused. Maybe this is just a stepping stone? 

This is definitely a stepping stone for server interceptors as a whole - it solves the global server interceptors path.
The service specific path I am handling next in #5. Sorry that I did not mention that in my description.

> But the docs don't seem to make sense for the non-global case - `ServerBuilder.intercept()` is a method to add a global interceptor and you use it in the example of a non-global, unless I'm reading it wrong somehow. I think you need to create and wrap a service definition with `ServerInterceptors`.

Great catch. The 2nd half of the doc is incorrect and I will remove until I do #5 next.

> Also, I know Spring Integration has global interceptors, so we probably copied the pattern from there? 

The annotation was how they did it in the gRPC ecosystem and also we use similar annotations in other projects so it seems like a standard way to go. 

> But there is also the pattern of a "registration bean" with an enabled flag in Spring Boot for servlets and filters. That would work here too, and then we wouldn't need an annotation, or the annotation could be a convenient way of marking the interceptor registration as "enabled".

Ok, let me take a look into that a bit more tomorrow before we push this through.

dsyer:
I added support for registering beans of type `ServerServiceDefnition` so that should be the way to fix your service-specific interceptor docs.

PR #49, opened 2024-11-09
Update JavaDocs for grpc-client classes

panic08:
I think this will be good for us, since java-doc has already been described for grpc-server classes, so I guess it's worth doing the same for grpc-client classes


dsyer:
CI failure probably means you need to fix something. Please squash and rebase before you push.

dsyer:
Still failing. Please run `mvnw install` locally before you push the fix.

panic08:
I think everything should be okay now

PR #50, opened 2024-11-11
Add support for service specific server interceptors

onobc:
> [!NOTE]
> @dsyer this is still a WIP (needs docs / tests / etc..) but wanted to get some eyes on it before continuing.

This commit introduces the `@GrpcService` annotation which can be used to associate one or more `ServerInterceptor` beans with a `BindableService`.

See #5

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/ApplicationContextBeanLookupUtils.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/ApplicationContextBeanLookupUtils.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.grpc.autoconfigure.server;
+
+import java.lang.annotation.Annotation;
+import java.util.LinkedHashMap;
+import java.util.List;
+
+import org.springframework.context.ApplicationContext;
+import org.springframework.core.annotation.Order;
+import org.springframework.util.Assert;
+
+/**
+ * Convenience methods performing bean lookups that are not currently provided by the
+ * standard Spring facilities.
+ *
+ * @author Chris Bono
+ */
+final class ApplicationContextBeanLookupUtils {
+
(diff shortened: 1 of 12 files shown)


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/DefaultGrpcServiceConfigurer.java
```diff
@@ -0,0 +1,82 @@
...
+import io.grpc.ServerServiceDefinition;
+import jakarta.annotation.PostConstruct;
+
+/**
+ * Default {@link GrpcServiceConfigurer} that binds and configures services with
+ * interceptors.
+ *
+ * @author Chris Bono
+ */
+public class DefaultGrpcServiceConfigurer implements GrpcServiceConfigurer {
```
onobc, line 37: Broke the "bind and configure" part of the discoverer into this.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/ApplicationContextBeanLookupUtils.java
```diff
@@ -0,0 +1,96 @@
...
+	 * FactoryBeans will get initialized in order to determine their object type.
+	 * @param applicationContext the application context
+	 * @param beanType the type of bean
+	 * @param annotationType the type of annotation
+	 * @param <B> type of bean
+	 * @param <A> type of annotation
+	 * @return a map of matching bean instances with the annotation instance (or null)
+	 * ordered according to their {@link Order} with annotation
+	 */
+	static <B, A extends Annotation> LinkedHashMap<B, A> getOrderedBeansWithAnnotation(
```
onobc, line 56: These methods give us ordering and filtering by annotation (which is not available OOTB w/ AppContext nor BeanFactoryUtils nor BeanFactoryAnnotationUtils).

dsyer:
The only thing I'm not 100% comfortable with is the new interface `GrpcServiceConfigurer` because it has an annotation type in its method signature. I prefer that I can call the interface with an object that I can easily create myself, and the annotation is only used as an implementation detail. I can see why you did it that way but I think it might need a tweak, possibly at the expense of yet another abstraction.

onobc:
> The only thing I'm not 100% comfortable with is the new interface `GrpcServiceConfigurer` because it has an annotation type in its method signature. I prefer that I can call the interface with an object that I can easily create myself, and the annotation is only used as an implementation detail. I can see why you did it that way but I think it might need a tweak, possibly at the expense of yet another abstraction.

I agree w/ this and I noticed the struggle when writing tests for this. Its a bit challenging to create an annotation instance on the fly and pass it in 😄 . I will find a way to pull the annotation out of the signature. Good feedback.

onobc:
@dsyer I still need to handle a few tests and docs but the pieces are almost in place. Just ran out of time. You can see the wrapper I created to specify additional service info.

dsyer:
Sounds good. The only question I had was if maybe some of the new abstractions (like the service info record) could live in core?

onobc:
> Sounds good. The only question I had was if maybe some of the new abstractions (like the service info record) could live in core?

That is going to happen in the next commit for sure. I too noticed this. Did not want to get too many things moving at once though.

PR #53, opened 2024-11-11
Add missing metadata for `spring.grpc.server.observation.enabled`

making:
so that IDE can suggest the property


onobc (approved):
Thanks for adding this @making

Issue #12, opened 2024-09-16
Support for shutdown grace period in netty clients


dsyer:

panic08:
Do I understand correctly that the purpose of this issue is to create a “life cycle” beans for a client? Which are already in place for the server

dsyer:
No, I don't think so. The lifecycle is already handled by the `DefaultGrpcChannelFactory` (it is a `DisposableBean`). But there is no explicit configuration support for graceful shutdown (waiting for ongoing requests to finish before shutdown). I believe it is probably supported in the client, but we don't expose it anywhere.

panic08:
> No, I don't think so. The lifecycle is already handled by the `DefaultGrpcChannelFactory` (it is a `DisposableBean`). But there is no explicit configuration support for graceful shutdown (waiting for ongoing requests to finish before shutdown). I believe it is probably supported in the client, but we don't expose it anywhere.

Yes, `DefaultGrpcChannelFactory` implements `DisosableBean` and overrides the destroy method which closes channels without any delay. I guess the point of the question is to allow the user to specify a delay(grace period) after which the channel shutdown?

Can I be assigned to this issue?

PR #54, opened 2024-11-12
Support for shutdown grace period in netty clients

panic08:
This PR refers to #12

Here we give opportunity to set a specific shutdown grace period in properties. If the customer channel is not fulfilled within this time, it will be shutdowned now.

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
@@ -16,10 +16,12 @@
 
 package org.springframework.grpc.client;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 
 import org.springframework.beans.factory.DisposableBean;
 
@@ -50,6 +52,8 @@ public class DefaultGrpcChannelFactory implements GrpcChannelFactory, Disposable
 
 	private VirtualTargets targets = VirtualTargets.DEFAULT;
 
+	private Duration shutdownGracePeriod;
+
 	public DefaultGrpcChannelFactory() {
 		this(List.of());
 	}
@@ -64,6 +68,10 @@ public class DefaultGrpcChannelFactory implements GrpcChannelFactory, Disposable
 
 	public void setCredentialsProvider(ChannelCredentialsProvider credentials) {
 		this.credentials = credentials;
+	}
+
+	public void setShutdownGracePeriod(Duration shutdownGracePeriod) {
+		this.shutdownGracePeriod = shutdownGracePeriod;
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


dsyer:
Looks good. Two questions (both could be deferred to future work):

1) Would it be a good idea to try and shutdown all the clients in parallel, rather than waiting for each one?

2) Couldn't each channel have its own timeout? It looks like the model you made allows it, but then ignores them. If it's too complicated we'd either need a new abstraction in core, or we could pull the global timeout up out of the `NamedChannel`.

panic08:
> Looks good. Two questions (both could be deferred to future work):
> 
> 
> 
> 1) Would it be a good idea to try and shutdown all the clients in parallel, rather than waiting for each one?
> 
> 
> 
> 2) Couldn't each channel have its own timeout? It looks like the model you made allows it, but then ignores them. If it's too complicated we'd either need a new abstraction in core, or we could pull the global timeout up out of the `NamedChannel`.

I think it's a good idea to develop in the future. I guess the grace period for the client implies a grace period for each channel. 

However, making a grace period for all channels simultaneously and for individual channels is a good idea that can be developed in the future.

Perhaps this can be put into two separate parameters as well

dsyer:
I'm a bit worried that leaving the timeout as a `NamedChannel` property makes it obligatory to implement individual timeouts at some point. So I'd like to know if it's awkward or impossible at least, in case we don't want to do it. I can have a think but if you have any insight please share it.

panic08:
We can now rename shutdownGracePeriod to shutdownGracePeriodPerRequest, thus ensuring backward compatibility and then we won't have to worry if (and I guess when) we want to do a parallel shutdown grace period

dsyer:
I don't understand what "per request" means. Isn't it "per channel" (in which case the fact that it is a property of a `NamedChannel` makes it redundant)?

panic08:
Yes, I apologize, I mean “perChannel”.

So the current shutdownGracePeriod that I created we will rename directly now to shutdownGracePeriodPerChannel. This will solve any problems that may arise if we decide to create a parallel shutdownGracePeriod

dsyer:
OK, then the name doesn't need to change (because it's a property of a channel already). I don't think that solves any problems. Still thinking. If you have any ideas about how to implement the "per channel" behaviour then feel free to try them out.

panic08:
I assume that by “per channel” we mean setting a specific shutdownGracePeriod separately for each channel? Is it possible to theriotically create some kind of settings class for channels to solve this issue.

We could overload the createChannel method in GrpcChannelFactory and add another argument to the method with the settings class for the channel. Then directly in the DefaultGrpcChannelFactory we could monitor if the setting for that channel is present with the shutdownGracePeriod field, and if so, we use it for that channel in the destroy method. If not, we use the default one set in properties (or if not in properties - which we set systematically).

dsyer:
This seems kind of stale now. It might be better to close it and start again?

panic08:
I guess it's worth updating the changes made in this PR and doing a rebase? Pending a decision though, how do we deal with this problem in the best way possible

dsyer:
@onobc has some changes planned to the channel factory. I think it might be best to regroup when they come through.

PR #55, opened 2024-11-14
Add description to additional metadata fields

onobc:
This commit adds a description for the observations and reflection enabled properties.

I noticed we were missing descriptions in these enabled properties that do not exist in the server properties.


spring-grpc-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -7,11 +7,16 @@
     },
     {
       "name": "spring.grpc.server.reflection.enabled",
-      "defaultValue": "true"
```
onobc, line 10: The `spring.grpc.server.port` property exists w/in the config props and has a description. It was only in here because its default value could not be deduced.

These, on the other hand, exist outside of the config props so there is no description for them.

PR #57, opened 2024-11-14
Move some components into core module

onobc:
This commit moves several components into the core module from the autoconfiguration module.
Anything that exists in the AC module will ultimately end up in Spring Boot project assuming we move into Spring Boot in the future.


spring-grpc-core/src/main/java/org/springframework/grpc/server/service/DefaultGrpcServiceConfigurer.java
```diff
@@ -44,8 +45,8 @@ public DefaultGrpcServiceConfigurer(ApplicationContext applicationContext) {
 		this.applicationContext = applicationContext;
 	}
 
-	@PostConstruct
```
onobc, line 47: Did not want to mess w/ bringing in jakarta-annotations-api in core (it was dep mgmt in autoconfigure module via SB).

spring-grpc-core/src/test/java/org/springframework/grpc/server/service/DefaultGrpcServiceConfigurerTests.java
```diff
@@ -224,55 +273,6 @@ void whenBlendInterceptorsTrueThenGlobalInterceptorsBlended() {
 
 	}
 
-	private void customizeContextAndRunServiceConfigurerWithServiceInfo(
```
onobc, line 227: Checkstyles made me do it :)

PR #59, opened 2024-11-15
Use specific bean types in autoconfiguration

onobc:
This commit uses the most specific bean type in the autoconfiguration to improve AOT compatibility.

As per the following restriction
```
As we cannot rely on the instance, make sure that the bean type is as precise as possible.
```
from https://docs.spring.io/spring-framework/reference/core/aot.html#aot.introduction


spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
```diff
@@ -36,4 +36,4 @@
 |spring.grpc.server.ssl.enabled |  | Whether to enable SSL support. Enabled automatically if "bundle" is provided unless specified otherwise.
 |spring.grpc.server.ssl.secure | `+++true+++` | Flag to indicate that client authentication is secure (i.e. certificates are checked). Do not set this to false in production.
 
-|===
+|===
```
onobc, line 39: Sorry, **[UNRELATED]** from the formatter and it keeps doing it so I wanted to just check it so I can quite rolling it back before commits.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientAutoConfiguration.java
```diff
@@ -44,7 +44,7 @@ public class GrpcClientAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean(GrpcChannelFactory.class)
 	public DefaultGrpcChannelFactory defaultGrpcChannelFactory(final List<GrpcChannelConfigurer> configurers,
-			ChannelCredentialsProvider credentials, GrpcClientProperties channels, SslBundles bundles) {
+			ChannelCredentialsProvider credentials, GrpcClientProperties channels, SslBundles ignored) {
```
onobc, line 47: Eventhough we don't have checkstyles enforced in autoconfigure module yet on the command line. I have them setup in my IDE so when I go to commit it tells me all about it. These were harmless enough that I just piggybacked them here. Let me know if this is an issue and I can do it in subsequent commit or subsequent PR.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientAutoConfiguration.java
```diff
@@ -113,7 +114,7 @@ static class NamedChannelVirtualTargets implements VirtualTargets {
 		@Override
 		public String getTarget(String authority) {
 			NamedChannel channel = this.channels.getChannel(authority);
-			return channels.getTarget(channel.getAddress());
+			return this.channels.getTarget(channel.getAddress());
```
onobc, line 117: Ditto https://github.com/spring-projects-experimental/spring-grpc/pull/59/files#r1843164573

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerObservationAutoConfiguration.java
```diff
@@ -21,21 +21,19 @@
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
 
-import io.grpc.ServerInterceptor;
 import io.micrometer.core.instrument.binder.grpc.ObservationGrpcServerInterceptor;
 import io.micrometer.observation.ObservationRegistry;
 
 @AutoConfiguration(
 		afterName = "org.springframework.boot.actuate.autoconfigure.observation.ObservationAutoConfiguration")
-@ConditionalOnClass(value = { ObservationRegistry.class, ObservationGrpcServerInterceptor.class })
```
onobc, line 30: Ditto https://github.com/spring-projects-experimental/spring-grpc/pull/59/files#r1843164573

Issue #56, opened 2024-11-14
Expose actuator health check over gRPC health check endpoint if available


dsyer:
There is a "standard" health check endpoint provided by grpc-services: https://github.com/grpc/grpc-java/blob/master/services/src/main/proto/grpc/health/v1/health.proto. We could maybe hook that into the actuator health check if it's available (i.e. Spring Boot Actuator is on the classpath).

PR #63, opened 2024-11-21
Add autoconfiguration for server-side health

onobc:
This commit autoconfigures the gRPC health service by providing the following beans:

- gRPC HealthStatusManager
- gRPC HealthService
- Actuator health check adapter (skeleton impl)

See #56

> [!NOTE]
> The next 2 PRs will extend this one as follows:
> - [ ] implement the actuator adapter
> - [ ] add client-side health autoconfiguration

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/pom.xml
+++ b/spring-grpc-spring-boot-autoconfigure/pom.xml
@@ -84,8 +84,16 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter</artifactId>
-			<scope>provided</scope>
+			<artifactId>spring-boot</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-autoconfigure</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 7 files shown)


spring-grpc-spring-boot-autoconfigure/pom.xml
```diff
@@ -84,8 +84,16 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter</artifactId>
```
onobc, line 87: The `provided` did not play well w/ `optional` actuator. I believe just adding in w/o the boot starter works well too.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerProperties.java
```diff
@@ -131,10 +135,67 @@ public void setMaxInboundMetadataSize(DataSize maxInboundMetadataSize) {
...
+
+		/**
+		 * Whether to adapt Actuator health checks into gRPC health checks.
+		 */
+		private Boolean enabled = true;
+
+		/**
+		 * List of Actuator health checks to adapt into gRPC health checks.
+		 */
+		private List<String> endpoints = new ArrayList<>();
```
onobc, line 179: This will be used in the next phase but I just added them in pre-emptively

Issue #52, opened 2024-11-11
Support for ClientInterceptor


dsyer:
Same as for the server side, but possibly complicated by the design of the channel factory.

vuhp:
I would like to take this task. Can you assign it to me?

onobc:
Hi @vuhp , just touching base on this issue. We wanted to get this in for the upcoming release and I was going to grab this one if you had not had a chance to work on it yet. No worries if you have not had time - I know we all get busy. There are also plenty of other issues to grab. 

Thanks,
Chris

PR #69, opened 2024-11-30
[WIP] Add client interceptors to channel factory

onobc:
> [!NOTE]
> This is a rough 1st pass to get buy-in before moving forward further

**Commit 1:** just moves the bean lookup util to the internal package so it can be shared w/ client package.

This commit adds support for global and channel-specific client interceptors.

See #52

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorsConfigurer.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorsConfigurer.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * */
+
+package org.springframework.grpc.client;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.grpc.internal.ApplicationContextBeanLookupUtils;
+
+import io.grpc.ClientInterceptor;
+import io.grpc.ManagedChannelBuilder;
+
+/**
+ * Configure a {@link ManagedChannelBuilder} with client interceptors.
+ *
+ * @author Chris Bono
+ */
+public class ClientInterceptorsConfigurer implements InitializingBean {
+
+	private final ApplicationContext applicationContext;
+
+	private List<ClientInterceptor> globalInterceptors;
+
+	public ClientInterceptorsConfigurer(ApplicationContext applicationContext) {
+		this.applicationContext = applicationContext;
+	}
+
+	/**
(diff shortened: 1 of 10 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
```diff
@@ -34,4 +37,16 @@ public interface GrpcChannelFactory {
...
+	/**
+	 * Creates a {@link ManagedChannelBuilder} for the given authority and the provided
+	 * client interceptors.
+	 * @param authority the target authority for the channel
+	 * @param interceptors the non-null list of interceptors to be applied to the channel
+	 * @param mergeWithGlobalInterceptors whether the provided interceptors should be
+	 * blended with the global interceptors.
+	 * @return a channel builder conifgured with the provided values
+	 */
+	ManagedChannelBuilder<?> createChannel(String authority, List<ClientInterceptor> interceptors,
```
onobc, line 49: I decided not doing a builder for just 1 more method. I think we will want to add one if we add 1 more method to this. Wdyt?

spring-grpc-test/src/main/java/org/springframework/grpc/test/InProcessGrpcServerFactoryAutoConfiguration.java
```diff
@@ -49,8 +52,14 @@ InProcessGrpcServerFactory grpcServerFactory(GrpcServiceDiscoverer grpcServicesD
 	}
 
 	@Bean
-	InProcessGrpcChannelFactory grpcChannelFactory() {
-		InProcessGrpcChannelFactory factory = new InProcessGrpcChannelFactory();
+	@ConditionalOnMissingBean
```
onobc, line 55: I don't like that this bean is repeated here. If we want to pass one into the channel factory we need to create it here or split the client auto-config to include a `@Configuration` that both could import (this AC has a before condition on client AC).

spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorsConfigurer.java
```diff
@@ -0,0 +1,77 @@
...
+
+import io.grpc.ClientInterceptor;
+import io.grpc.ManagedChannelBuilder;
+
+/**
+ * Configure a {@link ManagedChannelBuilder} with client interceptors.
+ *
+ * @author Chris Bono
+ */
+public class ClientInterceptorsConfigurer implements InitializingBean {
```
onobc, line 35: So this is not like the other `GrpcChannelConfigurer` but I could not find a better name yet. 

How do you feel about renaming the `GrpcChannelConfigurer` to `GrpcChannelBuilderCustomizer` like we did for the server customizers? Then this CIC would be the default ClientConfigurer? I am in favor of doing that.

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -46,16 +48,16 @@ public class DefaultGrpcChannelFactory implements GrpcChannelFactory, Disposable
...
 
 	private VirtualTargets targets = VirtualTargets.DEFAULT;
 
-	public DefaultGrpcChannelFactory() {
-		this(List.of());
-	}
-
-	public DefaultGrpcChannelFactory(List<GrpcChannelConfigurer> configurers) {
+	public DefaultGrpcChannelFactory(List<GrpcChannelConfigurer> configurers,
+			ClientInterceptorsConfigurer interceptorsConfigurer) {
```
onobc, line 58: I added to `<init>` as it is required if user wants any kind of interceptors. I see that you did `targets` and `credentials` w/ setters. I am guessing to avoid constructor bloat and possible to let users override if need be? Should we do the same for this CIC?

onobc:
Resubmitting PR in more logical commit order as things have changed a bit w/ the impl since this submission.

PR #71, opened 2024-12-01
Add client interceptors to channel factory

### LINKED ISSUE #52: Support for ClientInterceptor

onobc:
This commit adds support for global and channel-specific client interceptors.

Resolves #52

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
@@ -16,6 +16,9 @@
 
 package org.springframework.grpc.client;
 
+import java.util.List;
+
+import io.grpc.ClientInterceptor;
 import io.grpc.ManagedChannelBuilder;
 
 /**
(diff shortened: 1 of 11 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
```diff
@@ -34,4 +37,16 @@ public interface GrpcChannelFactory {
...
+	/**
+	 * Creates a {@link ManagedChannelBuilder} for the given authority and the provided
+	 * client interceptors.
+	 * @param authority the target authority for the channel
+	 * @param interceptors the non-null list of interceptors to be applied to the channel
+	 * @param mergeWithGlobalInterceptors whether the provided interceptors should be
+	 * blended with the global interceptors.
+	 * @return a channel builder conifgured with the provided values
+	 */
+	ManagedChannelBuilder<?> createChannel(String authority, List<ClientInterceptor> interceptors,
```
onobc, line 49: I decided not doing a builder for just 1 more method. I think we will want to add one if we add 1 more method to this. Wdyt?

spring-grpc-test/src/main/java/org/springframework/grpc/test/InProcessGrpcServerFactoryAutoConfiguration.java
```diff
@@ -49,8 +52,14 @@ InProcessGrpcServerFactory grpcServerFactory(GrpcServiceDiscoverer grpcServicesD
 	}
 
 	@Bean
-	InProcessGrpcChannelFactory grpcChannelFactory() {
-		InProcessGrpcChannelFactory factory = new InProcessGrpcChannelFactory();
+	@ConditionalOnMissingBean
+	ClientInterceptorsConfigurer inProcessClientInterceptorsConfigurer(ApplicationContext applicationContext) {
```
onobc, line 56: I don't like that this bean is repeated here. If we want to pass one into the channel factory we need to create it here or split the client auto-config to include a `@Configuration` that both could import (this AC has a before condition on client AC).

dsyer: Agree. Let’s do this soon

Issue #67, opened 2024-11-27
Virtual targets and duplicate entries in channels map


dsyer:
To support the `NamedChannel` feature in the Spring Boot jar we added an indirection in core between the logical name of a channel and its address. This is going to be a source of bugs, and we can see already that the default implementation of `GrpcChannelFactory` ends up with extra map entries when user configures a named channel (one for the logical name and one for the address). Maybe that was intentional, and maybe it doesn't matter, but we should check it out and add some more tests to make sure that if there are duplicates they both point to the same physical channel. It might be a good idea to move the `NamedChannel` metadata object into core (would also help with #12).

onobc:
A test can easily be added to `https://github.com/spring-projects-experimental/spring-grpc/commit/e0199dd13000d5e1d0bf185598f7b01758cb9cec#diff-b11d84d1b9f35c855a55cf4e126a1330265aaadef4dc7878dc6cfb7e9d1ae4edR223` to reproduce the extra map entry issue.

PR #72, opened 2024-12-01
Virtual targets and duplicate entries in channels map

panic08:
Refers to #67

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientAutoConfiguration.java
@@ -86,7 +86,7 @@ public class GrpcClientAutoConfiguration {
 		@Override
 		public String getTarget(String authority) {
 			NamedChannel channel = this.channels.getChannel(authority);
-			return this.channels.getTarget(channel.getAddress());
+			return this.channels.getTarget(channel);
 		}
 
 	}
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
@@ -83,6 +83,15 @@ public class GrpcClientProperties implements EnvironmentAware {
 		return address;
 	}
 
+	public String getTarget(NamedChannel channel) {
+		String address = channel.getAddress();
+		if (address.startsWith("static:") || address.startsWith("tcp:")) {
+			address = address.substring(address.indexOf(":") + 1).replaceFirst("/*", "");
+		}
+		address = this.environment.resolvePlaceholders(address);
+		return address;
+	}
+
 	public static class NamedChannel {
 
 		/**
--- a/spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/client/GrpcClientPropertiesTests.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/client/GrpcClientPropertiesTests.java
@@ -200,6 +200,19 @@ class GrpcClientPropertiesTests {
 			assertThat(properties.getChannels()).containsOnlyKeys("custom");
 		}
 
+		@Test
+		void withCustomChannelReturnsDuplicateEntryMap() {
+			Map<String, String> map = new HashMap<>();
+			map.put("spring.grpc.client.channels.custom.address", "static://my-server:8888");
+			GrpcClientProperties properties = bindProperties(map);
+			GrpcClientAutoConfiguration.NamedChannelVirtualTargets virtualTargets = new GrpcClientAutoConfiguration.NamedChannelVirtualTargets(
+					properties);
+			var address = virtualTargets.getTarget("custom");
+			assertThat(address).isEqualTo("my-server:8888");
+			assertThat(properties.getTarget("custom")).isEqualTo("my-server:8888");
+			assertThat(properties.getChannels()).containsOnlyKeys("custom");
+		}
+
 	}
 
 }


dsyer:
Looks like we need a rebase? Also maybe `NamedChannelCredentialsProvider` can move into core as well?

panic08:
> Looks like we need a rebase? Also maybe `NamedChannelCredentialsProvider` can move into core as well?

I'm not quite sure, but I don't think so. We can't use from core, for example `SslBundles` or `GrpcClientProperties`.

Also an interesting point. Right now the build is crashing for obvious reasons. If we look carefully in the log, we can see that: 
```
[ERROR] GrpcClientPropertiesTests$BindPropertiesAPI.withoutKeepAliveUnitsSpecified:125 
expected: 1S
 but was: 0.001S
```

This is due to the fact that from core `@DurationUnit` (which I removed from `NamedChannel` as it is not reachable) was responsible for converting from application.properties to a specific value (seconds). But now it is in milliseconds.

I guess we should think about how to play this properly to handle time correctly?

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientAutoConfiguration.java
```diff
@@ -25,14 +25,8 @@
...
 import org.springframework.context.annotation.Import;
-import org.springframework.grpc.autoconfigure.client.GrpcClientProperties.NamedChannel;
 import org.springframework.grpc.autoconfigure.common.codec.GrpcCodecConfiguration;
-import org.springframework.grpc.client.ChannelCredentialsProvider;
-import org.springframework.grpc.client.ClientInterceptorsConfigurer;
-import org.springframework.grpc.client.DefaultGrpcChannelFactory;
-import org.springframework.grpc.client.GrpcChannelBuilderCustomizer;
-import org.springframework.grpc.client.GrpcChannelFactory;
-import org.springframework.grpc.client.VirtualTargets;
+import org.springframework.grpc.client.*;
```
onobc, line 29: Please remove this wildcard import.

onobc:
> > Looks like we need a rebase? Also maybe `NamedChannelCredentialsProvider` can move into core as well?
> 
> I'm not quite sure, but I don't think so. We can't use from core, for example `SslBundles` or `GrpcClientProperties`.
> 
> Also an interesting point. Right now the build is crashing for obvious reasons. If we look carefully in the log, we can see that:
> 
> ```
> [ERROR] GrpcClientPropertiesTests$BindPropertiesAPI.withoutKeepAliveUnitsSpecified:125 
> expected: 1S
>  but was: 0.001S
> ```
> 
> This is due to the fact that from core `@DurationUnit` (which I removed from `NamedChannel` as it is not reachable) was responsible as I understand for converting from application.properties to a specific value (seconds) if no unit is specified. But now the default unit of measure is milliseconds.
> 
> I guess we should think about how to play this properly to handle time correctly?

I'm going to pull down and take a look closer. Ulitmately we will want the auto-configuration to map from client properties to the created named channels. The current named channel was doing both setting of defaults as well as representing channels. More likely I would expect a NamedChannel record that is constructed from the NamedChannel properties.

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -72,6 +72,10 @@ public void setCredentialsProvider(ChannelCredentialsProvider credentials) {
 
 	@Override
 	public ManagedChannelBuilder<?> createChannel(String authority) {
+		String target = this.targets.getTarget(authority);
```
onobc, line 75: So `this.targets.getTarget` ultimately goes into NamedVirtualTargets in the auto-config:
```java
		@Override
		public String getTarget(String authority) {
			NamedChannel channel = this.channels.getChannel(authority);
			return this.channels.getTarget(channel.getAddress());
		}
```
which ultimately still calls `this.channels.getChannel` and `this.channels.getTarget` which in turn still does the duplicate mapping. I don't think this code is doing what you think it is.

So far this PR seems to only be moving the NamedChannel out of the config props and into core.

onobc: I would prefer the first thing we do is surface the duplicate entries issue in a test and get that figured out and fixed and then look into moving the NamedChannel into core.

onobc: BTW - here is a rough prototype of how I would split the named channel out of autconfgure module. https://github.com/spring-projects-experimental/spring-grpc/pull/74

Note that something similar could be done for the NamedCredentials. 

@dsyer wdyt?

panic08:
I guess we should still move the NamedChannel migration to another PR then, and put the test writing and fixing into this one.

By the way, I don't know the project very well and I would like you to help me a bit. Am I right that this code can help us and be a test?

```
ClientInterceptorsConfigurer configurer = mock();

DefaultGrpcChannelFactory factory = new DefaultGrpcChannelFactory(List.of(), configurer);

ManagedChannel logicalChannel = factory.createChannel("logicalName").build();
ManagedChannel physicalChannel = factory.createChannel("static://localhost:9090").build();

assertThat(logicalChannel).isSameAs(physicalChannel);
```

onobc:
> I guess we should still move the NamedChannel migration to another PR then, 

I think that is a good approach. 

> and put the test writing and fixing into this one (or should we?).

I think so.

> 
> By the way, I don't know the project very well and I would like you to help me a bit. Am I right that this code can help us and be a test?
> 
> ```
> ClientInterceptorsConfigurer configurer = mock();
> 
> DefaultGrpcChannelFactory factory = new DefaultGrpcChannelFactory(List.of(), configurer);
> 
> ManagedChannel logicalChannel = factory.createChannel("logicalName").build();
> ManagedChannel physicalChannel = factory.createChannel("static://localhost:9090").build();
> 
> assertThat(logicalChannel).isSameAs(physicalChannel);
> ```

I will take a look at the above code sometime in the next few hours.

onobc:
Hi @panic08 ,

The particular "duplicate entries" can be seen when there is a named channel configured w/ an address as follows:
```
spring.grpc.client.channels.custom.address=static://0.0.0.0:0
```
Then a call to `DefaultGrpcChannelFactory.createChannel("custom")` results in an extra entry in the `NamedVirtualTargets` channels map.

Let's look at the call chain to understand what's going on...
```
DefaultGrpcChannelFactory.createChannel("custom")
	NamedChannelVirtualTargets.getTarget("custom")
		GrpcClientProperties.getChannel("custom")
		GrpcClientProperties.getTarget("static://0.0.0.0:0")
			GrpcClientProperties.getChannel("static://0.0.0.0:0")
			GrpcClientProperties.getTarget("static://0.0.0.0:0")
```

On line 4 above, you can see `getTarget` called w/ the actual address which results in the `GrpcClientProperties.getTarget` doing a lookup in the channels map (`computeIfAbsent`).

The main issue is that `NamedChannelVirtualTargets.getTarget` calls `GrpcClientProperties.getTarget` with the channel address which in turn goes back through the `getChannel+getTarget` with the address.

You can add the following test to `org.springframework.grpc.autoconfigure.client.GrpcClientPropertiesTests.GetTargetAPI`

```java
@Test
void withCustomChannelReturnsDuplicateEntryMap() {
	Map<String, String> map = new HashMap<>();
	map.put("spring.grpc.client.channels.custom.address", "static://my-server:8888");
	GrpcClientProperties properties = bindProperties(map);
	NamedChannelVirtualTargets virtualTargets = new NamedChannelVirtualTargets(properties);
	var address = virtualTargets.getTarget("custom");
	assertThat(address).isEqualTo("my-server:8888");
	assertThat(properties.getTarget("custom")).isEqualTo("my-server:8888");
	assertThat(properties.getChannels()).containsOnlyKeys("default", "custom");
}
```

I know it is a lot of details but just let me know if you need more clarification etc..

panic08:
Thank you very much for your clarification

In `NamedChannelVirtualTargets.getTarget` I have removed the double call and replaced with
```
@Override
public String getTarget(String authority) {
	return this.channels.getChannel(authority).getAddress();
}
```
and also added the test in `GrpcClientPropertiesTests.GetTargetAPI` you helped to write.

Also would like to understand. Am I correct in understanding that you made a slight mistake and instead of
`assertThat(properties.getTarget(“custom”)).isEqualTo(“my-server:8888”);`
should be
`assertThat(properties.getTarget(“custom”)).isEqualTo(“static://my-server:8888”);`

panic08:
Hmm. As I understand it, we can use '''GrpcClientProperties.getChannel''' to get the address. We also want to commit getTarget for the address, but without calling getChannel. I will think about how this can best be done.

onobc:
> Also would like to understand. Am I correct in understanding that you made a slight mistake and instead of

No it is intentional, as the `GrpcClientProperties.getTarget` strips the `static` prefix off:
```java
if (address.startsWith("static:") || address.startsWith("tcp:")) {
	address = address.substring(address.indexOf(":") + 1).replaceFirst("/*", "");
}
```

panic08:
I apologize for the inactivity of two weeks.

To solve this problem, I thought about it and decided to resort to overloading the `getTarget` method in `GrpcClientProperties` with the `NamedChannel` argument.

I also decided that it would make sense to change
`assertThat(properties.getChannels()).containsOnlyKeys(“default”, “custom”);`

which you recommended to
`assertThat(properties.getChannels()).containsOnlyKeys(“custom”);`

I did it because of the https://github.com/spring-projects-experimental/spring-grpc/commit/acbe334b7ee42520c4e6e380bea20a3595198185

onobc (approved):
Thanks for the contribution @panic08 - nice work.

PR #74, opened 2024-12-02
WIP: Split NamedChannel out of autoconfigure

onobc:


spring-grpc-core/src/main/java/org/springframework/grpc/client/NamedChannel.java
```diff
@@ -0,0 +1,13 @@
+package org.springframework.grpc.client;
+
+import java.time.Duration;
+
+public record NamedChannel(String address,
```
onobc, line 5: Of course, there will be more fields here (just a sample)

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientAutoConfiguration.java
```diff
@@ -48,14 +48,22 @@ ClientInterceptorsConfigurer clientInterceptorsConfigurer(ApplicationContext app
 		return new ClientInterceptorsConfigurer(applicationContext);
 	}
 
+	@Bean
+	NamedChannelRegistry namedChannelRegistry(GrpcClientProperties properties) {
```
onobc, line 52: This is the abstraction used in core now

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
```diff
@@ -39,23 +36,16 @@
 import io.grpc.netty.NettyChannelBuilder;
 
 @ConfigurationProperties(prefix = "spring.grpc.client")
-public class GrpcClientProperties implements EnvironmentAware {
+public class GrpcClientProperties {
```
onobc, line 39: Note we still keep all the config props named channel variant here which is needed for type-safe config props, auto-completion in IDE, etc.. We just map them to the NamecChannel records in core (registry)

onobc:
Closing in favor of https://github.com/spring-projects-experimental/spring-grpc/pull/89

PR #76, opened 2024-12-04
Fix config props defaults and descriptions

onobc:
This commit simplifies the client properties by setting the default values directly on the properties and removes markdown from the property javadocs. 
This improves end user experience by adding default values and descriptions to the generated config props docs.


spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/client/GrpcClientPropertiesTests.java
```diff
@@ -146,37 +146,10 @@ class CopyDefaultsAPI {
...
 			GrpcClientProperties properties = new GrpcClientProperties();
 			var defaultChannel = properties.getDefaultChannel();
-			var newChannel = new GrpcClientProperties.NamedChannel();
-			newChannel.copyDefaultsFrom(defaultChannel);
+			var newChannel = defaultChannel.copy();
 			assertThat(newChannel).usingRecursiveComparison().isEqualTo(defaultChannel);
 		}
 
-		@Test
-		void withUserSpecifiedValuesAreRetained() {
```
onobc, line 155: While this was previously supported w/ `copyDefaultsFrom` we were not using it.

PR #78, opened 2024-12-11
Improve GrpcChannelFactory API

onobc:
- Introduce a `ChannelBuilderOptions` that holds customizer, shutdownGracePeriod, and interceptors.
- Add type to GrpcChannelFactory and GrpcChannelBuilderCustomizer


spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+
+/**
+ * An extended {@link ManagedChannelBuilder} that allows setting the shutdown grace period
+ * for created channels as well as more options for client interceptors.
+ *
+ * @author Chris Bono
+ */
+public class GrpcChannelBuilder extends ForwardingChannelBuilder2<GrpcChannelBuilder> {
```
onobc, line 33: Just like the NettyServerBuilder provides extended functionality, we do the same here for:

1. Add shutdown grace period
2. Add merge interceptors indicator
3. Override existing intercept method append or merge w/ global interceptors depending on above indicators (NYI but it shows the contract and how we can do it)

spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
```diff
@@ -16,37 +16,33 @@
...
-	 * @param authority the target authority for the channel
-	 * @return a {@link ManagedChannelBuilder} configured for the given authority
-	 */
-	ManagedChannelBuilder<?> createChannel(String authority);
-
-	/**
-	 * Creates a {@link ManagedChannelBuilder} for the given authority and the provided
-	 * client interceptors.
-	 * @param authority the target authority for the channel
-	 * @param interceptors the non-null list of interceptors to be applied to the channel
```
onobc, line 44: This functionality is all in the GrpcChannelBuilder now.

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -71,68 +66,41 @@ public void setCredentialsProvider(ChannelCredentialsProvider credentials) {
...
-	}
-
-	@Override
-	public ManagedChannelBuilder<?> createChannel(String authority, List<ClientInterceptor> interceptors,
-			boolean mergeWithGlobalInterceptors) {
-		Assert.notNull(interceptors, () -> "interceptors must not be null");
-		return this.builders.computeIfAbsent(authority, path -> {
-			ManagedChannelBuilder<?> builder = newChannelBuilder(this.targets.getTarget(path),
-					this.credentials.getChannelCredentials(path));
-			this.interceptorsConfigurer.configureInterceptors(builder, interceptors, mergeWithGlobalInterceptors);
```
onobc, line 85: Will have to pass the interceptors configurer to the new builder to let it do its magic (or something like it)

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -71,68 +66,41 @@ public void setCredentialsProvider(ChannelCredentialsProvider credentials) {
...
 	@Override
-	public ManagedChannelBuilder<?> createChannel(String authority) {
-		return this.createChannel(authority, List.of(), false);
-	}
-
-	@Override
-	public ManagedChannelBuilder<?> createChannel(String authority, List<ClientInterceptor> interceptors,
-			boolean mergeWithGlobalInterceptors) {
-		Assert.notNull(interceptors, () -> "interceptors must not be null");
-		return this.builders.computeIfAbsent(authority, path -> {
```
onobc, line 82: Caching of builders is gone

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -71,68 +66,41 @@ public void setCredentialsProvider(ChannelCredentialsProvider credentials) {
...
 	 * @param creds the credentials for the channel
-	 * @return a new {@link ManagedChannelBuilder} for the given path and credentials
+	 * @return a new builder for the given path and credentials
 	 */
 	protected ManagedChannelBuilder<?> newChannelBuilder(String path, ChannelCredentials creds) {
 		return Grpc.newChannelBuilder(path, creds);
 	}
 
-	private ManagedChannel buildAndRegisterChannel(String channelName, ManagedChannelBuilder<?> channelBuilder) {
-		return this.channels.computeIfAbsent(channelName, (__) -> channelBuilder.build());
```
onobc, line 103: Caching of channels is gone as well (**for now**).

I recommend that we skip caching of the channels for now. Because we hand out the builders for post-handout mutations, we would need to be able to `Function<Builder, CacheKey>` and that is a bumpy road that I would like to holdoff traveling down.

We still handle the lifecycle (shutdown), but users have to cache the channels themselves.

Wdyt?

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -71,68 +66,41 @@ public void setCredentialsProvider(ChannelCredentialsProvider credentials) {
...
+		ManagedChannel channel = channelBuilder.delegate().build();
+		this.channels.add(new ManagedChannelWithShutdown(channel, channelBuilder.shutdownGracePeriod()));
+		return channel;
 	}
 
 	@Override
 	public void destroy() {
-		this.channels.values().forEach(ManagedChannel::shutdown);
+		this.channels.forEach((c) -> {
+			var shutdownGracePeriod = c.shutdownGracePeriod();
```
onobc, line 96: 🎉 `shutdown with grace`

onobc (commented):
@dsyer I believe this extended builder concept solves several of our current pain points.

> [!NOTE]
> I bring the following point up now because it could impact the new extended channel builder design in this code proposal.

I noticed that the `NettyChannelBuilder` static methods handle the `unix` scheme, for example:

```java
if (DOMAIN_SOCKET_ADDRESS_SCHEME.equals(address.getScheme())) {
    final String path = GrpcUtils.extractDomainSocketAddressPath(address.toString());
    return NettyChannelBuilder.forAddress(new DomainSocketAddress(path))
            .channelType(EpollDomainSocketChannel.class)
            .eventLoopGroup(new EpollEventLoopGroup());
} else {
    return NettyChannelBuilder.forTarget(address.toString())
            .defaultLoadBalancingPolicy(properties.getDefaultLoadBalancingPolicy());
}
```

1. Does the `Grpc.newChannelBuilder()` we are currently leveraging support `unix` scheme for Netty?
2. Refresh my memory why we did not have the `createChannel` return the specific `ManagedChannelBuilder` (e.g. `NettyChannelBuilder`)? Was it for simplicity?

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
dsyer: Caching of channels is still there in `this.channels`. I think that makes sense - not sure how we'd handle lifecycle without caching channels.

dsyer:
> 1. Does the `Grpc.newChannelBuilder()` we are currently leveraging support `unix` scheme for Netty?

At the time I wrote it I believed not (copied code from ecosystem), but looking at the `UdsNameResolverProvider` I'd say it probably does.

> 2. Refresh my memory why we did not have the `createChannel` return the specific `ManagedChannelBuilder` (e.g. `NettyChannelBuilder`)? Was it for simplicity?

I didn't want the user to have to know what kind of channel it is (same for servers). Mainly because I anticipated substituting the in-process version in tests. Maybe that was a bad decision - we'd have to support the in-process transport in a different way if we change the contract.

spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelBuilder.java
dsyer: I wanted to avoid this as it's so close to the native `ManagedChannelBuilder`, but I think you have proved it is probably necessary.

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
dsyer: Good.

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
onobc: **[Discussed offline]** summary: the channels are "remembered" so we can handle lifecycle but not "cached" in traditional sense from a user perspective (i.e. if a user asks for the same target 2x, they get 2x channels, not 1).

spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
```diff
@@ -16,37 +16,32 @@
...
+ * @author Chris Bono
  */
 public interface GrpcChannelFactory {
 
 	/**
-	 * Creates a {@link ManagedChannelBuilder} for the given authority.
-	 * @param authority the target authority for the channel
-	 * @return a {@link ManagedChannelBuilder} configured for the given authority
-	 */
-	ManagedChannelBuilder<?> createChannel(String authority);
```
dsyer, line 38: I think maybe we should keep this overload. It can have a default implementation.

spring-grpc-core/src/main/java/org/springframework/grpc/client/ChannelBuilderOptions.java
```diff
@@ -135,11 +135,13 @@ public ChannelBuilderOptions withShutdownGracePeriod(Duration shutdownGracePerio
 
 	/**
 	 * Set the customizer to apply to the builder used to create the channel.
+	 * @param <T> type of builder the customizer operates on
 	 * @param customizer the customizer to apply to the builder used to create the channel
 	 * @return a new immutable options instance populated with the specified
 	 * {@code customizer} and the settings of this current options instance.
 	 */
-	public ChannelBuilderOptions withCustomizer(GrpcChannelBuilderCustomizer<?> customizer) {
+	public <T extends ManagedChannelBuilder<T>> ChannelBuilderOptions withCustomizer(
```
onobc, line 143: This makes it more lambda-friendly (examples [here]())

dsyer:
LGTM. Maybe we should use the overloaded default factory method in the tests where possible though (no need for the second argument if it’s just the default)?

PR #81, opened 2024-12-17
Add autoconfiguration for client-side observation interceptor

### LINKED ISSUE #51: Support for client side observability
The server side works fine now via the autconfigured interceptor but there's nothing for clients yet. Probably best to wait for #52, or use this as a way to investigate design work for that issue.

NotBjoggisAtAll:
[resolves #51]

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/pages/client.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/pages/client.adoc
@@ -131,3 +131,8 @@ However, by setting the `mergeWithGlobalInterceptors` parameter on the channel f
 You can use this option if you want to add a per-client interceptor between global interceptors.
 
 IMPORTANT: The per-channel interceptors you pass in must either be bean instances marked with `@Order` or regular objects that implement the `Ordered` interface to be properly merged/ordered with the global interceptors.
+
+== Observability
+
+Spring gRPC provides an autoconfigured interceptor that can be used to provide observability to your gRPC services.
+All you need to do is add Spring Boot actuators to your project, and optionally a bridge to your observability platform of choice (just like https://docs.spring.io/spring-boot/reference/actuator/observability.html[any other Spring Boot application]).
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/pages/client.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/pages/client.adoc
@@ -134,5 +134,4 @@ IMPORTANT: The per-channel interceptors you pass in must either be bean instance
 
 == Observability
 
-Spring gRPC provides an autoconfigured interceptor that can be used to provide observability to your gRPC services.
-All you need to do is add Spring Boot actuators to your project, and optionally a bridge to your observability platform of choice (just like https://docs.spring.io/spring-boot/reference/actuator/observability.html[any other Spring Boot application]).
+Spring gRPC provides an autoconfigured interceptor that can be used to provide observability to your gRPC clients.
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientObservationAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientObservationAutoConfiguration.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.grpc.autoconfigure.client;
+
+import io.micrometer.core.instrument.binder.grpc.ObservationGrpcClientInterceptor;
+import io.micrometer.observation.ObservationRegistry;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.grpc.client.GlobalClientInterceptor;
+
+@AutoConfiguration(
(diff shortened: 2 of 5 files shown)


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientObservationAutoConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.grpc.client.GlobalClientInterceptor;
+
+@AutoConfiguration(
+		afterName = "org.springframework.boot.actuate.autoconfigure.observation.ObservationAutoConfiguration")
+@ConditionalOnClass({ ObservationRegistry.class, ObservationGrpcClientInterceptor.class })
+@ConditionalOnBean(ObservationRegistry.class)
+@ConditionalOnProperty(name = "spring.grpc.client.observation.enabled", havingValue = "true", matchIfMissing = true)
```
onobc, line 31: Do you mind adding an entry to the [config props additional metadata](https://github.com/spring-projects-experimental/spring-grpc/blob/729940d62c15fb198e5e3232fec7f06031f80ede/spring-grpc-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L21)?

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/client/GrpcClientObservationAutoConfigurationTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+	@Test
+	void whenObservationPropertyDisabledThenAutoConfigIsSkipped() {
+		this.validContextRunner()
+			.withPropertyValues("spring.grpc.client.observation.enabled=false")
+			.run(context -> assertThat(context).doesNotHaveBean(GrpcClientObservationAutoConfiguration.class));
+	}
+
+	@Test
+	void whenAllConditionsAreMetThenInterceptorConfiguredAsExpected() {
```
onobc, line 80: I like this test you have added. It checks if the interceptor bean is registered and if its marked as global interceptor. I think the test may be a bit easier to follow by using `hasEntrySatisifying` - something like:
```java
@Test
void whenAllConditionsAreMetThenInterceptorConfiguredAsExpected() {
this.contextRunner().run((context) -> {
	assertThat(context).hasSingleBean(ObservationGrpcClientInterceptor.class);
	assertThat(context.getBeansWithAnnotation(GlobalClientInterceptor.class))
			.hasEntrySatisfying("observationGrpcClientInterceptor", 
					(bean) -> bean.getClass().isAssignableFrom(ObservationGrpcClientInterceptor.class));
});
}
```
Wdyt?

onobc (changes requested):
Thanks for the contribution @NotBjoggisAtAll . nice work. I added a couple minor comments/suggestions, but other than that LGTM.

onobc:
I almost forgot, we have a small section on the server side observability docs [here](https://docs.spring.io/spring-grpc/reference/server.html#_observability).
It would be nice to do the same in the client section.
@NotBjoggisAtAll if you don't get to this that is ok - we can add it during merge etc..

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/client/GrpcClientObservationAutoConfigurationTest.java
NotBjoggisAtAll: Yes I agree. 

I copied the tests from [GrpcServerObservationAutoConfigurationTests](https://github.com/spring-projects-experimental/spring-grpc/blob/729940d62c15fb198e5e3232fec7f06031f80ede/spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerObservationAutoConfigurationTests.java).
 Do you want me to adjust the test for the server too? Should that be done in a separate PR?

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientObservationAutoConfiguration.java
NotBjoggisAtAll: Yes, I added it below the server version of this property

NotBjoggisAtAll:
I added an section for Observability for the client. I excluded the last part that was written on the server docs because I don't think we have a client sample yet?

I also had to adjust the test a bit. I think an assert inside the lamda is correct? Otherwise it seems that the test passes no matter what I put in `isAssignableFrom()`

Please let me know if there is anything else you want me to do

spring-grpc-docs/src/main/antora/modules/ROOT/pages/client.adoc
```diff
@@ -131,3 +131,8 @@ However, by setting the `mergeWithGlobalInterceptors` parameter on the channel f
 You can use this option if you want to add a per-client interceptor between global interceptors.
 
 IMPORTANT: The per-channel interceptors you pass in must either be bean instances marked with `@Order` or regular objects that implement the `Ordered` interface to be properly merged/ordered with the global interceptors.
+
+== Observability
+
+Spring gRPC provides an autoconfigured interceptor that can be used to provide observability to your gRPC services.
+All you need to do is add Spring Boot actuators to your project, and optionally a bridge to your observability platform of choice (just like https://docs.spring.io/spring-boot/reference/actuator/observability.html[any other Spring Boot application]).
```
dsyer, line 138: Would actuators be necessary or appropriate for a client app (no server)?

dsyer:
You need to build locally too, to make sure the CI build succeeds. It tells you how to fix it.

spring-grpc-docs/src/main/antora/modules/ROOT/pages/client.adoc
NotBjoggisAtAll: Yes I agree, I can remove the part talking about Spring Boot actuators and leave only the first line.

If you have more feedback on what else should be stated in the docs please let me know. Writing docs is not my strong side 😅

NotBjoggisAtAll:
> You need to build locally too, to make sure the CI build succeeds. It tells you how to fix it.

Yeah I noticed, I will take a look at it soon

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/client/GrpcClientObservationAutoConfigurationTest.java
onobc: Thats ok @NotBjoggisAtAll , I will update that one separately.

onobc (approved):

dsyer:
Merged in 4149ed5. Thank you.

PR #88, opened 2025-01-01
Disable in-process transport by default and add @AutoConfigureInProcessTransport annotation

### LINKED ISSUE #77: Test apps should not use in-process transport
We made the in-process transport the default for tests, but tests with `main()` methods want those to run with the actual TCP socket open. For now you have to manually add `spring.grpc.inprocess.enabled=false` to the main method, but it would be better if we found a way to flip the default for a non-test execution. Is there a way to detect that the application context is being created by spring-test?

panic08:
Here I have added the `AutoConfigureInProcessTransport` annotation which should be set over the test where we want in-process transport to be enabled. I also added the class `InProcessTransportEnvironmentPostProcessor` which implements `EnvironmentPostProcessor` and where I use `AnnotationUtils.findAnnotation` to check if the startup class is marked with an annotation or not. If yes, then we need to enable in-process transport and we set `spring.grpc.inprocess.enabled=true`. I also set `matchIfMissing` to false in `InProcessGrpcServerFactoryAutoConfiguration` (ignored as it is by default) so that in-process will default to false.

Closes #77

### DIFF
--- a/spring-grpc-test/src/main/java/org/springframework/grpc/test/AutoConfigureInProcessTransport.java
+++ b/spring-grpc-test/src/main/java/org/springframework/grpc/test/AutoConfigureInProcessTransport.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.grpc.test;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.boot.autoconfigure.ImportAutoConfiguration;
+
+/**
+ * Annotation that can be applied to a test class to enable auto-configuration for
+ * in-process transport.
+ *
+ * @author Andrey Litvitski
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@ImportAutoConfiguration
+public @interface AutoConfigureInProcessTransport {
+
+	boolean enabled() default true;
+
+}
--- a/samples/grpc-server-netty-shaded/src/test/java/com/example/demo/DemoApplicationTests.java
+++ b/samples/grpc-server-netty-shaded/src/test/java/com/example/demo/DemoApplicationTests.java
@@ -19,8 +19,7 @@ import com.example.demo.proto.HelloReply;
 import com.example.demo.proto.HelloRequest;
 import com.example.demo.proto.SimpleGrpc;
 
-@SpringBootTest(properties = { "spring.grpc.client.channels.test.address=static://localhost:0",
-		"spring.grpc.inprocess.enabled=false" })
+@SpringBootTest(properties = { "spring.grpc.client.channels.test.address=static://localhost:0" })
 public class DemoApplicationTests {
 
 	private static Log log = LogFactory.getLog(DemoApplicationTests.class);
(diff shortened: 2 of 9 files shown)


samples/grpc-tomcat/src/test/java/org/springframework/grpc/sample/ListenOnTwoPortsTests.java
```diff
@@ -15,11 +15,13 @@
 import org.springframework.grpc.sample.proto.HelloReply;
 import org.springframework.grpc.sample.proto.HelloRequest;
 import org.springframework.grpc.sample.proto.SimpleGrpc;
+import org.springframework.grpc.test.AutoConfigureInProcessTransport;
 import org.springframework.grpc.test.LocalGrpcPort;
 import org.springframework.test.annotation.DirtiesContext;
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
 		properties = { "spring.grpc.server.servlet.enabled=false", "spring.grpc.server.port=0" })
+@AutoConfigureInProcessTransport
```
dsyer, line 24: This is wrong? I think we need it to be off in this test.

samples/grpc-webflux/src/test/java/org/springframework/grpc/sample/GrpcServerApplicationTests.java
```diff
@@ -16,10 +16,12 @@
 import org.springframework.grpc.sample.proto.HelloReply;
 import org.springframework.grpc.sample.proto.HelloRequest;
 import org.springframework.grpc.sample.proto.SimpleGrpc;
+import org.springframework.grpc.test.AutoConfigureInProcessTransport;
 import org.springframework.grpc.test.LocalGrpcPort;
 import org.springframework.test.annotation.DirtiesContext;
 
 @SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, properties = "spring.grpc.server.reactive.enabled=false")
+@AutoConfigureInProcessTransport
```
dsyer, line 24: This is not intended to be an in-process test, I think?

spring-grpc-test/src/main/java/org/springframework/grpc/test/AutoConfigureInProcessTransport.java
```diff
@@ -0,0 +1,27 @@
...
+package org.springframework.grpc.test;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface AutoConfigureInProcessTransport {
```
dsyer, line 25: Isn't this supposed to be `@ImportAutoConfiguration`? Copy the annotations from one of the existing `@AutoConfigure*` annotations.

spring-grpc-test/src/main/java/org/springframework/grpc/test/InProcessTransportEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Collections;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.env.EnvironmentPostProcessor;
+import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.MapPropertySource;
+
+public class InProcessTransportEnvironmentPostProcessor implements EnvironmentPostProcessor {
```
dsyer, line 26: I'm not sure this is going to work. Don't we need something that is aware of the test environment? Look at the existing `@AutoConfigure*` annotations to see how they are implemented.

dsyer (changes requested):

spring-grpc-test/src/main/java/org/springframework/grpc/test/InProcessTransportEnvironmentPostProcessor.java
panic08: Thanks so much for the tip. I've looked into it more, tested it, and I'd like to say. Could we create an InProcessFactory that will inherit from ContextCustomizerFactory? That is, make a class like  [ObservabilityContextCustomizerFactory](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/actuate/observability/ObservabilityContextCustomizerFactory.java)

panic08:
I listened to what you said. For an idea for looking at `@AutoConfigure*` annotation implementations, I decided to go with [AutoConfigureObservability](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/actuate/observability/AutoConfigureObservability.java) as advised by @onobc . 

I created an `InProcessTransportContextCustomizerFactory` that will check for the presence of the `@AutoConfigureInProcessTransport` annotation I created. If this annotation is not present, I've added the ability for properties to specify `spring.grpc.inprocess.auto-configure` to enable auto-configuration (I think this will be good for us?). If auto-configure is enabled, for now we only do one thing, in the test properties set `spring.grpc.inprocess.enabled=true`

dsyer:
I took a version of your PR from yesterday and added some extra code to detect if we are running in a test. I think this works now. Merged in e0a744b. Thank you.

PR #94, opened 2025-01-12
Remove wrong example in documentation

DumiJDev:
Hi, I found a wrong example in documentation:

```java
@Bean
SimpleGrpc.SimpleBlockingStub stub(GrpcChannelFactory channels) {
	return SimpleGrpc.newBlockingStub(channels.createChannel("local").build());
}
```

Method build doesn't exist in createChannel, and is unnecessary.


onobc:
Thanks for the fix @DumiJDev . This was my bad as I missed these doc updates in b08e11ad393e9fccd3beed2136867ccfe715c9d3. I also found a couple other ones to fix and did that in a follow-on commit fc1de937a8b5c1337a780ea2b1d9d2aec5ad4097. 

Closing in favor of b861d9301b4ef232ef79a6bea1896d6c254ac72a

PR #95, opened 2025-01-13
Add graceful shutdown of channels

### LINKED ISSUE #12: Support for shutdown grace period in netty clients

onobc:
This commit provides graceful shutdown of all channels created by the channel factory.

Resolves #12

### DIFF
--- a/spring-grpc-core/src/test/java/org/springframework/grpc/client/DefaultGrpcChannelFactoryShutdownTests.java
+++ b/spring-grpc-core/src/test/java/org/springframework/grpc/client/DefaultGrpcChannelFactoryShutdownTests.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


spring-grpc-core/src/test/java/org/springframework/grpc/client/DefaultGrpcChannelFactoryShutdownTests.java
```diff
@@ -0,0 +1,126 @@
...
+		// create channels using factory and options
+		var c1 = channelFactory.createChannel("c1",
+				ChannelBuilderOptions.defaults().withShutdownGracePeriod(Duration.ofSeconds(7)));
+		var c2 = channelFactory.createChannel("c2",
+				ChannelBuilderOptions.defaults().withShutdownGracePeriod(Duration.ofSeconds(5)));
+		var c3 = channelFactory.createChannel("c3",
+				ChannelBuilderOptions.defaults().withShutdownGracePeriod(Duration.ofSeconds(10)));
+
+		// spy each channel to wait accordingly
+		var c1Spy = setupSpy(c1);
```
onobc, line 70: I am not excited about having to get in and spy like this. However, it is a necessary evil to get some basic test coverage.

Namely, we ensure each channel gets the following:
- `shutdown` invoked
- `awaitTermination` invoked
- `shutdownNow` never invoked **EXCEPT** only on channel2 when `exceedAwaitTime = true`

PR #96, opened 2025-01-13
Move named channel back into autoconfigure module

### LINKED ISSUE #92: Fix generated config props for named channels
I need to look into why this is not finding the default any longer. I have an idea but you can ignore this part of the code for the review for now.

_Originally posted by @onobc in https://github.com/spring-projects-experimental/spring-grpc/pull/89#discussion_r1903359674_

onobc:
This commit moves the channel config ("NamedChannel") back to autoconfigure module as the effort to move to core turns out to be more trouble than it is worth at this time.

Fixes #92


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
```diff
@@ -15,35 +15,415 @@
...
+		String address = name;
+		if (!name.contains(":/") && !name.startsWith("unix:")) {
+			address = "static://" + name;
+		}
+		channel.setAddress(address);
+		return channel;
+	}
+
+	@Override
+	public String getTarget(String authority) {
```
onobc, line 93: I found that having the getTarget next to the getChannel made the code a bit easier to follow than when it was in the NamedChannelVirtualTargets.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
```diff
@@ -15,35 +15,415 @@
...
+		if (address.startsWith("static:") || address.startsWith("tcp:")) {
+			address = address.substring(address.indexOf(":") + 1).replaceFirst("/*", "");
+		}
+		return this.environment.resolvePlaceholders(address);
+	}
+
+	/**
+	 * Represents the configuration for a {@link ManagedChannel gRPC channel}.
+	 */
+	public static class ChannelConfig {
```
onobc, line 105: This is identically the NamedChannel from core renamed to ChannelConfig. Removing the "named" from the name removes the implication that it is has a name but rather it is just channel configuration. 

I am ok w/ `NamedChannel` as well though and really I know better than renaming during a move around like this. I am glad to move it back - just lmk.

PR #102, opened 2025-01-24
Fix documentation on testing server application

### LINKED ISSUE #101: Incorrect documentation on testing server application
The [documentation on testing](https://docs.spring.io/spring-grpc/reference/server.html#_testing) states that all that is required is to include a dependency on `spring-grpc-test`.

> Testing
> If you include spring-grpc-test in your project, your gRPC server in a @SpringBootTest will be started in-process (i.e. not listening on a network port). All clients that connect to any server via the autoconfigured GrpcChannelFactory will be able to connect to it. You can switch the in-process server off by setting spring.grpc.in-process.enabled to false.

However, in order to get this to work, one has to set the property `spring.grpc-inprocess.enabled` to `true` explicitly. It is not enabled by default, as can be seen [here](https://github.com/spring-projects-experimental/spring-grpc/blob/main/spring-grpc-test/src/main/java/org/springframework/grpc/test/InProcessGrpcServerFactoryAutoConfiguration.java#L38):

```java
 @ConditionalOnProperty(prefix = "spring.grpc.inprocess", name = "enabled", havingValue = "true")
```

Also, the property has no hyphen between `in` and `process`, as stated in the documentation.

panic08:
Fixes #101

The main problem is that inprocess has recently had to be specified manually (we've been working on this here #88).

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/pages/server.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/pages/server.adoc
@@ -160,9 +160,9 @@ A `GrpcExceptionHandler` can be used to handle exceptions of a specific type, re
 
 == Testing
 
-If you include `spring-grpc-test` in your project, your gRPC server in a `@SpringBootTest` will be started in-process (i.e. not listening on a network port).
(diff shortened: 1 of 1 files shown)


dsyer:
Per your comment in the related issue, the docs are still wrong right? They say that the "in-process" transport "will be used", implying it is automatic, but you need to set the enabled property explicitly. It's a tiny thing, but if you are going to send a PR we might as well make it complete.

PR #106, opened 2025-02-02
Add fine-grained starter modules

onobc:
This PR does the following:

- Makes `grpc-netty` optional and leaves the opinion to the starter
- Cleans up dependencies so that no opinion on choice of server is made until the starter modules
- Adds the following fine-grained starters:
  - Netty client starter
  - Netty server starter
  - Servlet server starter
- Current coarse-grained starter points to newly added client + server starters
- Updates Tomcat samples to use new servlet server starter

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,7 @@
 		<module>spring-grpc-spring-boot-autoconfigure</module>
 		<module>spring-grpc-spring-boot-starter</module>
 		<module>spring-grpc-client-spring-boot-starter</module>
+		<module>spring-grpc-server-spring-boot-starter</module>
 		<module>spring-grpc-server-web-spring-boot-starter</module>
 		<module>samples</module>
 	</modules>
--- a/pom.xml
+++ b/pom.xml
@@ -20,8 +20,8 @@
 		<module>spring-grpc-test</module>
 		<module>spring-grpc-spring-boot-autoconfigure</module>
 		<module>spring-grpc-spring-boot-starter</module>
-        <module>spring-grpc-client-spring-boot-starter</module>
-        <module>spring-grpc-server-web-spring-boot-starter</module>
+		<module>spring-grpc-client-spring-boot-starter</module>
+		<module>spring-grpc-server-web-spring-boot-starter</module>
 		<module>samples</module>
 	</modules>
 
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,7 @@
 		<module>spring-grpc-spring-boot-autoconfigure</module>
 		<module>spring-grpc-spring-boot-starter</module>
         <module>spring-grpc-client-spring-boot-starter</module>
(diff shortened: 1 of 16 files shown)


.github/workflows/ci-pr.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           ./mvnw -B clean install
       - name: Capture Test Results
         if: failure()
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
onobc, line 28: **[UNRELATED]** Github started requiring usage of next version of the action.

pom.xml
```diff
@@ -20,6 +20,7 @@
 		<module>spring-grpc-test</module>
 		<module>spring-grpc-spring-boot-autoconfigure</module>
 		<module>spring-grpc-spring-boot-starter</module>
+        <module>spring-grpc-client-spring-boot-starter</module>
```
onobc, line 23: There is something off w/ my xml format settings. I will get these fixed here in this PR and push another commit.

spring-grpc-client-spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
onobc, line 2: > [!NOTE]
> Comment is general but put here as it is easier to reply than in a summary comment

Wrt to` start.spring.io,` we currently provide one option that includes server and client (`spring-grpc-spring-boot-starter`). I think we should wait until we are somewhat firm on the individual starters before making changes in Initializr as that is pretty outward facing and also may get pushback from team as this is still "experimental". 

If you agree, I will then make another starter `spring-grpc-server-spring-boot-starter` (w/ exact contents of the current `spring-grpc-spring-boot-starter`) and then just make the current`spring-grpc-spring-boot-starter` include both client and server starters.

From an outward perspective this is non-breaking change and we can refine what we advertise in start.spring.io at a later date. 

Wdyt?

onobc: I went ahead and added the above in [last commit](https://github.com/spring-projects-experimental/spring-grpc/pull/106/commits/67cde8601c399425d6b2d1db770197952cfa9ecc). I think it is the way to go. We can always drop it out of the PR if you disagree though.

dsyer:
This looks ready. Conflicts in the pesky config props doc. Does it make sense to turn the generation of that file into a manual step, so it doesn't pollute all our commits with bogus changes?

onobc:
> This looks ready. Conflicts in the pesky config props doc. Does it make sense to turn the generation of that file into a manual step, so it doesn't pollute all our commits with bogus changes?

I captured this here https://github.com/spring-projects-experimental/spring-grpc/issues/109

PR #108, opened 2025-02-08
Add config prop metadata and remove not-needed property

onobc:
- Adds a couple missing entries to additional-spring-configuration-metadata.json.

- Removes not-needed conditional property guard for `spring.grpc.server.reactive.enabled`

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -25,7 +25,7 @@ jobs:
           ./mvnw -B clean install
       - name: Capture Test Results
         if: failure()
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: test-results
           path: '*/target/surefire-reports/*.*'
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-pr.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           ./mvnw -B clean install
       - name: Capture Test Results
         if: failure()
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
onobc, line 28: **[UNRELATED]** Github started requiring usage of next version of the action.

spring-grpc-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -1,6 +1,12 @@
 {
   "groups": [],
   "properties": [
+    {
+      "name": "spring.grpc.security.csrf.enabled",
```
onobc, line 5: I think this should be named `spring.grpc.server.security.csrf.enabled` wdyt?

PR #110, opened 2025-02-13
Generate config props during doc phase

### LINKED ISSUE #109: Generate config props during doc phase
Currently the config props adoc is generated at normal build time (i.e. `mvn install` or `mvn package`). In some conditions, the generated doc will have blank values. This happens for @dsyer quite often. 

This results in toil of having to figure out why the file changed as well as a source of merge conflicts.

This file rarely changes so it would be nice to remove it out of the normal build workflow to avoid the above toil.

Let's make it part of the docs build.

therepanic:
I suppose the main problem is that we should not generate documentation at `mvn:install/package`, so if I assume correctly, then we should delegate documentation generation to `mvn:site`. Then the documentation will be generated only when `mvn:site` is activated.

Fix #109

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,7 +35,7 @@ jobs:
         echo BUILD_VERSION=$BUILD_VERSION >> $GITHUB_ENV
     - name: Run Antora
       run: |
-        ./mvnw -pl spring-grpc-docs antora
+        ./mvnw -pl spring-grpc-docs process-resources antora -P docs
     - name: Publish Docs
       uses: spring-io/spring-doc-actions/rsync-antora-reference@v0.0.11
       with:
--- a/spring-grpc-docs/pom.xml
+++ b/spring-grpc-docs/pom.xml
@@ -100,30 +100,6 @@
 				</configuration>
 			</plugin>
 			<plugin>
-				<groupId>org.codehaus.mojo</groupId>
-				<artifactId>exec-maven-plugin</artifactId>
-				<version>${maven-exec-plugin.version}</version>
-				<executions>
-					<execution>
-						<id>generate-configprops</id>
-						<phase>package</phase>
-						<goals>
-							<goal>java</goal>
-						</goals>
-						<configuration>
-							<includeProjectDependencies>true</includeProjectDependencies>
-							<includePluginDependencies>false</includePluginDependencies>
-							<mainClass>
-								org.springframework.grpc.internal.ConfigurationPropertiesAsciidocGenerator</mainClass>
-							<arguments>
-								<argument>${configprops.path}</argument>
-								<argument>${configprops.inclusionPattern}</argument>
-							</arguments>
-						</configuration>
-					</execution>
-				</executions>
-			</plugin>
-			<plugin>
 				<groupId>org.jruby.maven</groupId>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -222,7 +222,7 @@
 						<goals>
 							<goal>jar</goal>
 						</goals>
-						<phase>package</phase>
+						<phase>site</phase>
```
onobc, line 225: It is not the javadocs that we are wanting to avoid generating, but rather the config props which is controlled [here](https://github.com/spring-projects-experimental/spring-grpc/blob/8d20f4a6d918579616d39cb834f12757d225b1bd/spring-grpc-docs/pom.xml#L109). 

We do not generate our docs until the `./mvnw -pl spring-grpc-docs antora` command is run ([docs.yml](https://github.com/spring-projects-experimental/spring-grpc/blob/8d20f4a6d918579616d39cb834f12757d225b1bd/.github/workflows/docs.yml#L38)). 

We would like to make this config props generator run at that time. 

If this were Gradle it would be super simple to create a custom task named "generateConfigProps" (or something like that). With Maven it is not as simple.

panic08:
I apologize for not fully understanding the problem, I thought props were generated together with java docs, I got confused.

What I understood is that we want to generate the configuration props for the docs along with the command that is called when the docs are deployed ([here](https://github.com/spring-projects-experimental/spring-grpc/blob/8d20f4a6d918579616d39cb834f12757d225b1bd/.github/workflows/docs.yml#L38)). Couldn't we just call [this plugin](https://github.com/spring-projects-experimental/spring-grpc/blob/8d20f4a6d918579616d39cb834f12757d225b1bd/spring-grpc-docs/pom.xml#L109) in the “process-resources” phase, and assign a profile for it, let's say docs. Then, according to my logic, we should be able to generate configuration props and documentation together.

Let's say we can call the creation of documentation like this `./mvnw -pl spring-grpc-docs antora -Pdocs`

onobc:
> I apologize for not fully understanding the problem, I thought props were generated together with java docs, I got confused.

No worries. Thank you so much for all your contributions and jumping on these issues. 

> 
> What I understood is that we want to generate the configuration props for the docs along with the command that is called when the docs are deployed ([here](https://github.com/spring-projects-experimental/spring-grpc/blob/8d20f4a6d918579616d39cb834f12757d225b1bd/.github/workflows/docs.yml#L38)). Couldn't we just call [this plugin](https://github.com/spring-projects-experimental/spring-grpc/blob/8d20f4a6d918579616d39cb834f12757d225b1bd/spring-grpc-docs/pom.xml#L109) in the “process-resources” phase, and assign a profile for it, let's say docs. Then, according to my logic, we should be able to generate configuration props and documentation together.
> 
> Let's say we can call the creation of documentation like this `./mvnw -pl spring-grpc-docs process-resources antora -P docs`

Perfect! I almost forgot about Maven profiles. This is the way.

spencergibb:
This is how spring cloud does it with a docs profile

onobc (approved):
Nice work @panic08 . Thanks for another awesome contribution!

onobc:
@panic08 I also added a convenience script to build the docs and updated the docs README a bit. https://github.com/spring-projects-experimental/spring-grpc/commit/581e8415505b8a2519a227b7148b6f73b2bc16e1

PR #111, opened 2025-02-13
Fix typo in pom.xml in core module

panic08:
I'm not quite sure if `domgrpcn` makes any sense, so I assumed there was a typo here. I guess in our context `module` would be fine?


onobc (approved):
Good catch. Thanks @panic08

PR #119, opened 2025-02-28
Add spring-grpc-bom module

### LINKED ISSUE #118: Extract spring framework and security versions into build time config
We saw in #117 that users can experience issues with mismatched versions of Boot and Framework (and Security) because gRPC manages their versions independently, so anyone using out `spring-grpc-dependencies` BOM and `spring-boot-dependencies` at the same time can have a nasty surprise.

Probably the best solution is to pull those explicit versions out of our public dependencies BOM and put them in a separate configuration that is either not published at all, or at least not documented, so users can't easily include them by mistake.

onobc:
This adds a BOM module that includes the dependencies used by the framework except those that are managed by Spring Boot dependency management. The idea is that Spring Boot applications can use the BOM to get the proper versions for Spring gRPC and non-Spring Boot applications can use `spring-grpc-dependencies`.

Resolves #118

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,8 +14,6 @@
 	<description>Building gRPC applications with Spring Boot</description>
 
 	<modules>
-		<module>spring-grpc-docs</module>
-		<module>spring-grpc-dependencies</module>
 		<module>spring-grpc-core</module>
 		<module>spring-grpc-test</module>
 		<module>spring-grpc-spring-boot-autoconfigure</module>
@@ -23,6 +21,9 @@
 		<module>spring-grpc-client-spring-boot-starter</module>
 		<module>spring-grpc-server-spring-boot-starter</module>
 		<module>spring-grpc-server-web-spring-boot-starter</module>
+		<module>spring-grpc-bom</module>
+		<module>spring-grpc-dependencies</module>
+		<module>spring-grpc-docs</module>
 		<module>samples</module>
 	</modules>
 
--- a/spring-grpc-bom/pom.xml
+++ b/spring-grpc-bom/pom.xml
@@ -5,12 +5,12 @@
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>org.springframework.grpc</groupId>
-	<artifactId>spring-grpc-dependencies</artifactId>
+	<artifactId>spring-grpc-bom</artifactId>
 	<version>0.4.0-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
-	<name>Spring gRPC dependencies</name>
-	<description>Dependencies for the Spring gRPC modules</description>
+	<name>Spring gRPC (Bill of Materials)</name>
+	<description>Bill of Materials for the Spring gRPC modules</description>
 
 	<url>https://github.com/spring-projects-experimental/spring-grpc</url>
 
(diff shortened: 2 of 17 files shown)


pom.xml
```diff
@@ -14,15 +14,16 @@
...
-		<module>spring-grpc-docs</module>
-		<module>spring-grpc-dependencies</module>
 		<module>spring-grpc-core</module>
 		<module>spring-grpc-test</module>
 		<module>spring-grpc-spring-boot-autoconfigure</module>
 		<module>spring-grpc-spring-boot-starter</module>
 		<module>spring-grpc-client-spring-boot-starter</module>
 		<module>spring-grpc-server-spring-boot-starter</module>
 		<module>spring-grpc-server-web-spring-boot-starter</module>
+		<module>spring-grpc-bom</module>
```
onobc, line 24: Just moved these to the end as they are logically "after" the other modules.

spring-grpc-bom/pom.xml
```diff
@@ -0,0 +1,210 @@
...
+			</dependency>
+			<dependency>
+				<groupId>com.google.api.grpc</groupId>
+				<artifactId>proto-google-common-protos</artifactId>
+				<version>${google-common-protos.version}</version>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
+
+	<build>
```
onobc, line 122: I added the format plugin here w/ it set to skip as this was already an issue trying to run format command from the project root. The issue being that the parent of the "bom" and "dependencies" module is not the root pom and therefore did not have the plugin.

Longer term it may make sense (to help w/ DRY) to have a common parent for all the things that sets up the boilerplate pieces (somewhat like the `spring-cloud-build-parent`.

spring-grpc-docs/src/main/antora/modules/ROOT/pages/getting-started.adoc
```diff
@@ -185,11 +185,18 @@ repositories {
 [[dependency-management]]
 === Dependency Management
 
-The Spring gRPC Dependencies declares the recommended versions of all the dependencies used by a given release of Spring gRPC.
-Using the dependencies from your application’s build script avoids the need for you to specify and maintain the dependency versions yourself.
-Instead, the version you’re using determines the utilized dependency versions.
+The `spring-grpc-bom` artifact declares the recommended versions of the dependencies used by a given release of Spring gRPC, excluding dependencies already managed by Spring Boot dependency management.
```
spencergibb, line 188: I'm curious why bom for boot, dependencies for not boot? In spring cloud we use -dependencies for boot.

onobc: Great question @spencergibb , 

I find that more typically the "-bom" is for the modules provided by the project (and any other libs it uses that are very dependent on what it provides). E.g. `spring-grpc-bom` provides all of its modules/versions and also io.grpc and protobuf versions.

Whereas the "-dependencies" typically provide everything the "-bom" does as well as other libs not necessarily dependent on what it provides.

Some examples..

- https://mvnrepository.com/artifact/org.springframework/spring-framework-bom/6.2.3
- https://mvnrepository.com/artifact/org.springframework.amqp/spring-amqp-bom/3.2.3
- https://mvnrepository.com/artifact/org.springframework.integration/spring-integration-bom/6.4.2

And if you look at the whole list of "BOM" modules in Maven you will see a similar pattern

- https://mvnrepository.com/open-source/bom?p=1

In this case, we really have "BOM w/ spring boot managed deps" and "BOM w/o spring boot managed deps" and if we keep the names "<foo>-bom" and "<foo>-dependencies" I chose to go w/ "spring-grpc-bom" as the latter mostly based on pattern of what I see other modules doing. I see "-bom" as lighter than "-depedencies" usually.

onobc: @spencergibb me and @dsyer talked about this and we landed on:

- BOM = spring-grpc-dependencies
- ALL = spring-grpc-build-dependencies (BOM + SB managed deps)

Dave, I think you suggested `spring-grpc-build` (vs. `spring-grpc-build-dependencies`) but that seemed a bit generic to me. I almost want to name it `spring-grpc-dependencies-all` but that is a bit non-standard.

Spencer, Dave naming thoughts?

dsyer:
merged in b8ff6ea

PR #126, opened 2025-03-03
Add Spring Boot compatibility check workflow

onobc:
Adds a GHA workflow to check Spring Boot compatability of the framework.

The workflow exercises the samples and dynamically sets the version of each samples Spring Boot plugin.

We do the dynamic version setting using the power of Gradle init scripts and the fact that each of the samples is exposed as a Gradle module. Doing this has a positive side effect that our samples are kept up to date for both Maven and Gradle.

Reworks the Gradle aspect of the samples as follows:

- removes settings.xml in each sample in favor of root samples/setting.xml
- removes Gradle wrapper from each sample in favor of root samples/gradlew
- cleans up build.gradle (they were missing dependencies to work)
- **TODO** adds README-GRADLE.md at samples/


samples/settings.gradle
```diff
@@ -0,0 +1,17 @@
+pluginManagement {
+	repositories {
+		maven { url 'https://repo.spring.io/milestone' }
+		maven { url 'https://repo.spring.io/snapshot' }
+		gradlePluginPortal()
+	}
+}
+
+include 'grpc-reactive'
```
onobc, line 9: > [!NOTE]
> The `grpc-oauth2` has some Maven libs for testing and I just left out for now (we can add later)

dsyer: It’s just a test. There’s a new release of testjars that bundles them all together too, so we drop them soon.

PR #133, opened 2025-03-07
Add missing grpc-oauth2 submodule to samples/settings.gradle

panic08:
In #120 I was going to suggest adding HELP-.md to each submodule because they support native images, so I decided to build a grpc-oauth2 module for curiosity's sake. It didn't work out and as it turns out we forgot to include it in settings.gradle. Everything is fine in pom.xml

### DIFF
--- a/samples/settings.gradle
+++ b/samples/settings.gradle
@@ -6,6 +6,7 @@ pluginManagement {
 	}
 }
 
+include 'grpc-oauth2'
 include 'grpc-reactive'
 include 'grpc-secure'
 include 'grpc-server'


onobc (commented):
Good catch @panic08 , however this one is intentional due to [this](https://github.com/spring-projects-experimental/spring-grpc/pull/126/files#r1978013579).

dsyer:
> Good catch @panic08 , however this one is intentional due to [this](https://github.com/spring-projects-experimental/spring-grpc/pull/126/files#r1978013579).

That's not relevant any more since there was a testjars release. But the testjars dependency still needs to be added to the build.gradle, so this PR is incomplete.

dsyer:
Superseded in e71bdcf

PR #163, opened 2025-05-02
Remove redundant gradle wrappers in grpc-client sample

therepanic:
I think, similar to the rest of the samples, we should remove the gradle wrappers, since there are already wrappers on the outside of this sample


panic08:
I think that when we were testing this sample, we accidentally forgot to remove the wrappers.

PR #172, opened 2025-05-15
Use BlockingStubFactory by default if factory isn't specified in @ImportGrpcClients

genuss:
Due to a recent change https://github.com/spring-projects/spring-grpc/commit/8f706665df0a33c4dfb147126590f032488f2d9e the `@ImportGrpcClients` annotation requires specifying a factory. If it's not specified, [ReflectionUtils.findMethod](https://github.com/spring-projects/spring-grpc/blob/0720e8f647767d19de9ee08c0e2fa9e4ba298c8a/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcClientFactory.java#L187) will fail with an NPE.
This commit fixes this by using BlockingStubFactrory by default, just like it's done in DefaultGrpcClientRegistrations and written in javadoc for the annotation.

### DIFF
--- a/samples/grpc-server/src/test/java/org/springframework/grpc/sample/GrpcClientApplicationTests.java
+++ b/samples/grpc-server/src/test/java/org/springframework/grpc/sample/GrpcClientApplicationTests.java
@@ -67,12 +67,13 @@ public class GrpcClientApplicationTests {
 			assertThat(context.containsBeanDefinition("simpleFutureStub")).isTrue();
 			assertThat(context.getBean(SimpleGrpc.SimpleFutureStub.class)).isNotNull();
 			assertThat(context.containsBeanDefinition("simpleStub")).isFalse();
-			assertThat(context.containsBeanDefinition("simpleBlockingStub")).isFalse();
-			assertThat(context.getBeanNamesForType(AbstractStub.class)).hasSize(1);
+			assertThat(context.containsBeanDefinition("simpleBlockingStub")).isTrue();
+			assertThat(context.getBeanNamesForType(AbstractStub.class)).hasSize(2);
 		}
 
 		@TestConfiguration
 		@ImportGrpcClients(basePackageClasses = SimpleGrpc.class, factory = FutureStubFactory.class)
+		@ImportGrpcClients(basePackageClasses = SimpleGrpc.class)
 		static class TestConfig {
 
 		}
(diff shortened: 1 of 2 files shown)


dsyer:
I guess that's probably OK. I'm open to suggestion if anyone has other ideas. Unfortunately, you need to rebase and force push to make the CI build go green.

dsyer:
I actually don't think this will work yet. The integration tests in the grpc-secure sample fail because if there is no classpath scan then the factory has to be located automatically.

dsyer:
I implemented a different fix and made sure all the tests pass in 7de96df - the key is that the factory only has to fallback to a default if there is a classpath scan (explicitly listed types can match their own "natural" factory). Thanks for helping out.

genuss:
At the the same time I was checking out a similar fix. I also came up with an idea to register all stubs found by classpath scanner which can be created by available factories:
```diff
Index: spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcClientFactory.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcClientFactory.java b/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcClientFactory.java
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcClientFactory.java	(revision 822d229924904b7bf0830723ec566d37b91e70da)
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcClientFactory.java	(date 1747381479312)
@@ -264,6 +264,9 @@
 					@Override
 					public boolean match(MetadataReader metadataReader, MetadataReaderFactory metadataReaderFactory)
 							throws IOException {
+						if (GrpcClientRegistrationSpec.this.factory == null) {
+							return true;
+						}
 						Class<?> type = ClassUtils.resolveClassName(metadataReader.getClassMetadata().getClassName(),
 								ClasspathScanner.class.getClassLoader());
 						return supports(GrpcClientRegistrationSpec.this.factory, type);
```

It might though create too much beans, which won't ever be used, so your solution seems better.

dsyer:
Yeah, there was an explicit requirement *not* to create all possible beans on a scan (after some discussion on the original issue that motivated the feature). Hopefully we have the happy medium now.

Issue #170, opened 2025-05-14
Kotlin support


genuss:
Hi, team!
Let me express my appreciation for this project. I'm glad to see that spring-grpc is becoming part of the official Spring portfolio. I've been following this project for quite some time, and to be honest - I like it more and more. Congratulations on its promotion to a full member!

I'm interested, if there are plans for a better kotlin support. Having tried some basics I found #164 by myself and here. Another issue I found is that kotlin coroutines stubs aren't supported due to lack of a proper `StubFactory`. I created it and wanted to contribute, but I'd like to ask your opinion on the shape of my change. Putting the new factory in `StubFactory`'s package means that kotlin dependencies must be added to `spring-grpc-core` artifacts. I guess it's not desirable at least at the moment. That's why I'm asking a couple of questions:
* Do you plan for better kotlin support?
* How would you implement it: a separate artifact like `spring-grpc-kotlin`, a number of artifacts for client and server, or directly in `spring-grpc-core` with optional kotlin dependencies?

With these questions answered it would be possible to kickstart some development. I've already got a couple of features on my mind: coroutine stubs, `suspend`ing exception handlers, etc.

dsyer:
Better Kotlin support would be nice. You can look at some other projects (Spring Framework, Spring Boot, Spring Cloud Function probably the best since it is a Maven build) to see how Kotlin support has been added - probably a separate jar file makes sense for the core features at least. It's not a big thing, but finding people to contribute and maintain Kotlin features is often a challenge. So send a PR and then stick around, please, if you can.

dsyer:
Thanks. Please keep sending new features. If we need a new jar file at some point we can do it that way.

PR #173, opened 2025-05-15
Implement CoroutineStubFactory

genuss:
Following up #170 

Here is an implementation of `CoroutineStubFactory` useful for kotlin projects. First, I [implemented](https://github.com/genuss/spring-grpc/tree/coroutine-stub-factory-separtate-module) it in a separate module, as I was proposing in the original issue #170. However I it seems like an overkill at the moment. The only additional dependency is `grpc-kotlin-stub` which may safely be optional just like `reactor-grpc-stub`.

The factory is created as a bean if `grpc-kotlin-stub` is on a classpath. The alternative would be to put it into `GrpcClientFactory#DEFAULT_FACTORIES`, but I don't feel working around classpath issues in that class. Please correct me if you feel it's better to put it there.

I feel, I'd better add a `RuntimeHintsRegistrar` for creating coroutine stubs, but I could find one for standard ones. Is it actually done somewhere now?

### DIFF
--- a/spring-grpc-core/pom.xml
+++ b/spring-grpc-core/pom.xml
@@ -94,6 +94,11 @@
 			<artifactId>proto-google-common-protos</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-kotlin-stub</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>com.salesforce.servicelibs</groupId>
 			<artifactId>reactor-grpc-stub</artifactId>
 			<optional>true</optional>
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/CoroutineStubFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/CoroutineStubFactory.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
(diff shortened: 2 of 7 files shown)


dsyer:
Sorry, I think you have to rebase and force push to make the CI go green (I was messing about with the actions).

genuss:
Sure, it's done

dsyer:
Can you fix the failed code style checks? The import order one is a pain, but you can figure it out by looking at other similar classes (e.g. there has to be a new line before the org.springframework imports).

genuss:
Tried my best. Intellij does like the import style even with spring's formatting plugin.

dsyer:
```
[ERROR] /home/dsyer/dev/spring/spring-grpc/spring-grpc-core/src/test/java/org/springframework/grpc/client/GrpcClientFactoryTests.java:37:1: Wrong order for 'io.grpc.kotlin.AbstractCoroutineStub' import. [ImportOrder]
[ERROR] /home/dsyer/dev/spring/spring-grpc/spring-grpc-core/src/test/java/org/springframework/grpc/client/GrpcClientFactoryTests.java:151:17: Redundant 'public' modifier. [RedundantModifier]
```

Those 2 errors are not IDE specific (and anyway the majority of Spring developers use IntelliJ so we'd hear a lot about it if it was antagonistic against the code style). I'm pretty sure they are self explanatory. Anyway I fixed it for you in 6403bd9. Thanks.

Issue #184, opened 2025-05-23
Updating the 'Getting Started' section of README.md for gradle users


asingh04:
I noticed some things in the `README.md` could be fixed or updated in the `Getting Started` section for devs using gradle.

1.  Adding a comment to update the `java_package` name in the `hello.proto` file 
    -  _Not gradle related but felt it would be helpful to point out that simple copy/pasting would not work_
2. Include the proper path of the generated source directories for gradle
3. Fix the `bootRun` command

PR #185, opened 2025-05-23
fix: update README.md file for gradle instructions

asingh04:
Issue: #184 

Updating the `README.md`'s `Getting Started` section for users/dev using gradle.


onobc (approved):
Thanks for the contribution @asingh04 . It seems like a small detail but this will improve every Gradle users experience a bit when it comes to "Getting started". Much appreciated.

PR #187, opened 2025-05-26
WIP: Allow in-process server and channel outside of test module

onobc:


spring-grpc-test/src/main/java/org/springframework/grpc/test/InProcessGrpcServerFactoryAutoConfiguration.java
```diff
@@ -27,6 +27,8 @@
 import org.springframework.grpc.autoconfigure.client.GrpcClientAutoConfiguration;
 import org.springframework.grpc.autoconfigure.server.GrpcServerFactoryAutoConfiguration;
 import org.springframework.grpc.client.ClientInterceptorsConfigurer;
+import org.springframework.grpc.client.InProcessGrpcChannelFactory;
```
onobc, line 30: The usage of this in the test module has not been worked through yet - please ignore for now.

spring-grpc-core/src/main/java/org/springframework/grpc/client/InProcessOrAlternativeChannelFactory.java
```diff
@@ -0,0 +1,71 @@
...
+ * <li><code>in-process:foo/bar</code> -&gt; will use the <code>foo/bar</code> in-process-channel.</li>
+ * <li><code>static://127.0.0.1</code> -&gt; will be handled by the alternative grpc channel factory.</li>
+ * </ul>
+ *
+ * <p>
+ * Using this class does not incur any additional performance or resource costs, as the actual channels (in-process or
+ * other) are only created on demand.
+ * </p>
+ */
+public class InProcessOrAlternativeChannelFactory implements GrpcChannelFactory {
```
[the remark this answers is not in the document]
spencergibb: Why not do a true composite with an interface that has supports(String target)? We also use ordering so the default has lowest priority 
We do that in various places in cloud.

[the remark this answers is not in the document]
onobc: cc: @spencergibb @dsyer 

I went ahead and squashed/rebased as things have changed quite a bit.

Here is the gist though...

This update to the code proposal makes it so that we support...

**Server**
- A shaded Netty OR a Netty server factory (backing off if a user brings their own 1 or more server factories)
- An InProcess server factory if in-process is on the classpath and a name has been defined

**Client**
- A shaded Netty OR a Netty channel factory (backing off if a user brings their own 1 or more channel factories)
- An InProcess channel factory if in-process is on the classpath
- A composite channel factory composed of the above

**Test**
- An InProcess server factory that overrides the AC'ed one to use the spring.grpc.test.enabled=true driven address (i.e. InProcessServerBuilder.generateName())
- An InProcess channel factory that overrides the AC'ed one to always support any target (regardless of 'inprocess:' prefix) as well as a virtual targets function that echoes to use the spring.grpc.test.enabled=true driven address (i.e. InProcessServerBuilder.generateName())


> [!IMPORTANT]
> This update to the proposal does **NOT** solve the 'how do we tell which global client interceptors to apply to which channel builder' condundrum. Me and @dsyer have come up w/ an approach to use a `BiFunction<ManagedChannelBuilder, ClientInterceptor, Boolean>` with a good set of defaults but ultimately backing off when users have an opinion of their own. However, that will come in a subsequent commit once we are good on this current focus (enough moving pieces already).

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -70,6 +70,18 @@ public DefaultGrpcChannelFactory(List<GrpcChannelBuilderCustomizer<T>> globalCus
...
+	/**
+	 * Whether this factory supports the given target string. The target can be either a
+	 * valid nameresolver-compliant URI, an authority string as described in
+	 * {@link Grpc#newChannelBuilder(String, ChannelCredentials)}.
+	 * @param target the target string as described in method javadocs
+	 * @return true unless the target begins with 'in-process:'
+	 */
+	@Override
+	public boolean supports(String target) {
+		return !target.startsWith("in-process:");
```
onobc, line 82: All "default" channel factories (including those users may specialize from) support by default. I am not super excited about bleeding this here, however, it is a string and nothing more.

spring-grpc-core/src/main/java/org/springframework/grpc/client/GlobalClientInterceptor.java
```diff
@@ -42,4 +42,9 @@
 @Documented
 public @interface GlobalClientInterceptor {
 
+	/**
+	 * Whether to apply the interceptor to the {@link InProcessGrpcChannelFactory in-process} channel factory.
+	 * @return whether to apply the interceptor the in-process channel factory
+	 */
+	boolean includeInProcess() default true;
```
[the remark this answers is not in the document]
onobc: > [!NOTE]
> This will be addressed later via `BiFunction<ManagedChannelBuilder, ClientInterceptor, Boolean>`

spring-grpc-core/src/main/java/org/springframework/grpc/client/InProcessGrpcChannelFactory.java
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Construct an in-process channel factory instance.
+	 * @param globalCustomizers the global customizers to apply to all created channels
+	 * @param interceptorsConfigurer configures the client interceptors on the created
+	 * channels
+	 */
+	public InProcessGrpcChannelFactory(List<GrpcChannelBuilderCustomizer<InProcessChannelBuilder>> globalCustomizers,
+			ClientInterceptorsConfigurer interceptorsConfigurer) {
+		super(globalCustomizers, interceptorsConfigurer);
+		// TODO ???? setVirtualTargets((p) -> p);
```
onobc, line 40: I think it does make sense for the IPGCF to use an identity function by default for its virtual targets and will likely update this in next commit (realized I left in as a TODO as I did my self-pr-review).

spring-grpc-core/src/main/java/org/springframework/grpc/client/GlobalClientInterceptor.java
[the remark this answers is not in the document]
onobc: Agreed. I think its gone in the latest commit.

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
dsyer, line 82: I wonder if we should use "directaddress" instead of "in-process"? It's what gRPC calls it natively.

onobc: Unfortunately, the `InProcessChannelBuilder` will not strip that prefix off if its already there. So we would end up still having to do the stripping and then the name `directaddress` is further from the name of the feature. Because of this, I vote to keep it as-is. Wdyt?

dsyer:
I merged. You can add the docs when you feel like it.

onobc:
> I merged. You can add the docs when you feel like it.

Thank you @dsyer . Funny thing is I did the docs but looks like I did not push it upstream to this PR 😸 . Will follow up shortly/
Screenshot 2025-06-09 at 10 15 02

Issue #195, opened 2025-06-09
Add ability to filter out client interceptors for InProcess channel factories


onobc:
The main feature request from #144 was solved in  a087eac7764a22b9d89ee766283ddb0f17272e08. However, one aspect that we left for later was the ability to filter out the global client interceptors from the InProcess channel factories ([example here](https://github.com/spring-projects/spring-grpc/issues/144#issuecomment-2775694756)).

To prevent coupling the `@GlobalClientInterceptor` from the list of supported interceptors, an idea is to allow a `BiFunction<ClientInterceptor, GrpcChannelFactory, Boolean>` to be user-configured.

onobc:
Hi @Mejuh, this captures the other part of your request. However, we do not plan on adding this right now. If you need this ability though, just add a contribution and we will get it reviewed and merged in.

Thanks

therepanic:
Hi, @onobc! I think what we should do is that the user can create a bean with `BiFunction<ClientInterceptor, GrpcChannelFactory, Boolean>` and we in `ClientInterceptorsConfigurer` can find this bean and filter out the interceptors we don't need. I think I got the idea right? If nobody minds I can implement it and you can assign me.

onobc:
Hi @therepanic !
That is exactly what we had in mind. We would absolutely ❤ for you to work on this. Ping me if you have any questions, etc..

dsyer:
IMO we need an interface that isn't `BiFunction` so we can detect beans of that type more easily.

PR #196, opened 2025-06-15
Add support for filtering client interceptors in channel factories via user-configured filter bean

therepanic:
In this change, I added the ability for users to add filtering for `ClientInterceptor`.

I added the `ClientInterceptorFilter` interface, which contains the `shouldInclude` method, whose behavior will be configured by the users themselves when creating a bean. During the `afterPropertiesSet` event in `ClientInterceptorsConfigurer`, we find a bean with the new `ClientInterceptorFilter` interface and add all elements to the immutable list `List<ClientInterceptorFilter> interceptorFilters`.

Next, I extended the `configureInterceptors` method by adding a new argument `GrpcChannelFactory factory`. I made the current method an overload of the new one. Now, when this method is called, if `factory != null`, we remove all interceptors that did not pass the filter from `allInterceptors`. The current method was extended because some current tests test this method but do not use `GrpcChannelFactory`, but more on that later. I also added a test that tests this behavior, and everything passes.

However, there are some points that I am unsure about and would like to discuss with you:

1. Should we extend the current method? I implemented this behavior because some current methods do not mock `GrpcChannelFactory` and do not use it. I also did this because I thought that this method could possibly be called with other `GrpcChannelFactory`s. However, we can only have one `GrpcChannelFactory` bean in our system. And if we were to take this bean, we wouldn't have to extend anything. However, I decided to extend it anyway. Please let me know what you think about this.

2. I made a test that performs a simple mechanism: we register the `ClientInterceptorFilter` bean and filter all interceptors that are not `GlobalClientInterceptorsConfig.GLOBAL_INTERCEPTOR_BAR`. This test works correctly; if we didn't use filtering, the test would fail. However, do we want to add another test that checks that the test fails? Let's say we do something like this:
```java
var expected = List.of(GlobalClientInterceptorsConfig.GLOBAL_INTERCEPTOR_FOO, GlobalClientInterceptorsConfig.GLOBAL_INTERCEPTOR_BAR);
verify(builder).intercept(expected);
```
Then we would know for sure that everything works. If we run this test, it will work as intended (it will fail). Obviously, we want to check that it fails, but not that it fails on its own. But there is a nuance, and I don't know how to do it. That is, I tried something like this:
```java
assertThrows(ArgumentsAreDifferent.class, () -> {
    verify(builder).intercept(expected);
});
```
However, an error will be returned:
```java
java: cannot access junit.framework.ComparisonFailure
  class file for junit.framework.ComparisonFailure not found
```

Resolves: #195

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorFilter.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorFilter.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.grpc.client;
+
+import io.grpc.ClientInterceptor;
+
+/**
+ * Strategy to determine whether a {@link ClientInterceptor} should be included for a
+ * given {@link GrpcChannelFactory}.
+ *
+ * @author Andrey Litvitski
+ */
+@FunctionalInterface
+public interface ClientInterceptorFilter {
+
+	/**
+	 * Determine whether the given {@link ClientInterceptor} should be included for the
+	 * provided {@link GrpcChannelFactory}.
+	 * @param interceptor the client interceptor under consideration.
+	 * @param channelFactory the channel factory in use.
+	 * @return {@code true} if the interceptor should be included; {@code false}
+	 * otherwise.
+	 */
+	boolean filter(ClientInterceptor interceptor, GrpcChannelFactory channelFactory);
+
+}
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorsConfigurer.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorsConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024-2024 the original author or authors.
+ * Copyright 2024-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.List;
 
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.grpc.internal.ApplicationContextBeanLookupUtils;
 
@@ -31,12 +32,15 @@ import io.grpc.ManagedChannelBuilder;
  * Configure a {@link ManagedChannelBuilder} with client interceptors.
  *
  * @author Chris Bono
+ * @author Andrey Litvitski
  */
 public class ClientInterceptorsConfigurer implements InitializingBean {
 
 	private final ApplicationContext applicationContext;
 
 	private List<ClientInterceptor> globalInterceptors;
+
+	private ClientInterceptorFilter interceptorFilter;
 
 	public ClientInterceptorsConfigurer(ApplicationContext applicationContext) {
 		this.applicationContext = applicationContext;
@@ -48,13 +52,18 @@ public class ClientInterceptorsConfigurer implements InitializingBean {
 	 * @param interceptors the non-null list of interceptors to be applied to the channel
 	 * @param mergeWithGlobalInterceptors whether the provided interceptors should be
 	 * blended with the global interceptors.
+	 * @param factory the channel factory used to filter global interceptors
 	 */
 	protected void configureInterceptors(ManagedChannelBuilder<?> builder, List<ClientInterceptor> interceptors,
-			boolean mergeWithGlobalInterceptors) {
+			boolean mergeWithGlobalInterceptors, GrpcChannelFactory factory) {
 		// Add global interceptors first
 		List<ClientInterceptor> allInterceptors = new ArrayList<>(this.globalInterceptors);
 		// Add specific interceptors
 		allInterceptors.addAll(interceptors);
+		// Filter all interceptors
+		if (this.interceptorFilter != null) {
+			allInterceptors.removeIf(interceptor -> !this.interceptorFilter.filter(interceptor, factory));
+		}
 		if (mergeWithGlobalInterceptors) {
 			ApplicationContextBeanLookupUtils.sortBeansIncludingOrderAnnotation(this.applicationContext,
 					ClientInterceptor.class, allInterceptors);
@@ -66,11 +75,21 @@ public class ClientInterceptorsConfigurer implements InitializingBean {
 	@Override
 	public void afterPropertiesSet() {
 		this.globalInterceptors = findGlobalInterceptors();
+		this.interceptorFilter = findInterceptorFilter();
 	}
 
 	private List<ClientInterceptor> findGlobalInterceptors() {
 		return ApplicationContextBeanLookupUtils.getBeansWithAnnotation(this.applicationContext,
 				ClientInterceptor.class, GlobalClientInterceptor.class);
+	}
+
+	private ClientInterceptorFilter findInterceptorFilter() {
+		try {
+			return this.applicationContext.getBean(ClientInterceptorFilter.class);
+		}
+		catch (NoSuchBeanDefinitionException ignored) {
+			return null;
+		}
 	}
 
 }
--- a/spring-grpc-core/src/test/java/org/springframework/grpc/client/ClientInterceptorsConfigurerTests.java
+++ b/spring-grpc-core/src/test/java/org/springframework/grpc/client/ClientInterceptorsConfigurerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -43,6 +43,7 @@ import io.grpc.ManagedChannelBuilder;
  * Tests for {@link ClientInterceptorsConfigurer}.
  *
  * @author Chris Bono
+ * @author Andrey Litvitski
  */
 class ClientInterceptorsConfigurerTests {
 
@@ -67,8 +68,9 @@ class ClientInterceptorsConfigurerTests {
 			List<ClientInterceptor> clientSpecificInterceptors, List<ClientInterceptor> expectedInterceptors) {
 		ManagedChannelBuilder<?> builder = Mockito.mock();
 		this.contextRunner().with(contextCustomizer).run((context) -> {
+			var factory = Mockito.mock(GrpcChannelFactory.class);
 			var configurer = context.getBean(ClientInterceptorsConfigurer.class);
-			configurer.configureInterceptors(builder, clientSpecificInterceptors, true);
+			configurer.configureInterceptors(builder, clientSpecificInterceptors, true, factory);
 			// NOTE: the interceptors are called in reverse order per builder contract
 			var expectedInterceptorsReversed = new ArrayList<>(expectedInterceptors);
 			Collections.reverse(expectedInterceptorsReversed);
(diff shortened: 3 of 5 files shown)


therepanic:
I also assume that we should prepare documentation for this? If everything is correct, I could try to do it in separate commit in this PR.

dsyer:
It seems like the decision about whether to include interceptors doesn't need to be made until the channel is actually created? In that case we should add the target (url String) as an argument. Maybe the decision could even be delegated to the channel factory itself, at the time that it creates the channel? Then the `ClientInterceptorFilter` doesn't need to know about the channel factory.

therepanic:
So you are proposing to replace the argument in the new interface `channelFactory` with `target` and implement filtering in `createChannel`? I think this is also a good idea.

dsyer:
It's just an idea. If it doesn't lead to any method signature changes (apart from the new ones here) I think it might work well.

therepanic:
In this case, I believe we would like to have a `clientInterceptorFilters` field in `DefaultGrpcChannelFactory`. However, could you please advise how we can implement this bean in `GrpcChannelFactoryConfigurations` when creating a particular `GrpcChannelFactory`?

When creating a `GrpcChannelFactory` bean, can we add an additional `ApplicationContext` as argument for it? And do something like:
```java
@Bean
InProcessGrpcChannelFactory inProcessGrpcChannelFactory(ChannelBuilderCustomizers channelBuilderCustomizers,
		ClientInterceptorsConfigurer interceptorsConfigurer, ApplicationContext applicationContext) {
	List<GrpcChannelBuilderCustomizer<InProcessChannelBuilder>> inProcessBuilderCustomizers = List
		.of(channelBuilderCustomizers::customize);
	return new InProcessGrpcChannelFactory(inProcessBuilderCustomizers, interceptorsConfigurer,
			List.copyOf(applicationContext.getBeansOfType(ClientInterceptorFilter.class).values()));
}
```
In the case of the `testInProcessGrpcChannelFactory` bean, we will also have to do the following:
```java
@Bean
@Order(Ordered.HIGHEST_PRECEDENCE)
TestInProcessGrpcChannelFactory testInProcessGrpcChannelFactory(
		ClientInterceptorsConfigurer interceptorsConfigurer, ApplicationContext applicationContext) {
	return new TestInProcessGrpcChannelFactory(address, interceptorsConfigurer, List.copyOf(applicationContext.getBeansOfType(ClientInterceptorFilter.class).values()));
}
```

therepanic:
I also think that we will still have to overload `ClientInterceptorsConfigurer#configureInterceptors` and add `clientInterceptorFilters` and `target` as additional arguments.

onobc:
Hi @therepanic and @dsyer - I intended to review this by now but got busy w/ something else. I am going to review sometime today and will catch up on the comments etc..

onobc:
> I also assume that we should prepare documentation for this? If everything is correct, I could try to do it in separate commit in this PR.

I would do it in a separate commit and only after the main points are addressed (i.e. we know it will not change drastically and therefore have to re-do the docs).

onobc:
> Should we extend the current method?

I am in favor of not overloading the method. In this state of the project and due to to the fact it is a protected API we can just require a non-null factory. 

>  However, we can only have one GrpcChannelFactory bean in our system.

This is no longer true w/ the changes made in https://github.com/spring-projects/spring-grpc/commit/a087eac7764a22b9d89ee766283ddb0f17272e08. We do configure a primary composite channel factory if there are multiple non-primary channel factories configured. We should always have a single injectable primary channel factory though - but just wanted to clarify this.

onobc:
>  It seems like the decision about whether to include interceptors doesn't need to be made until the channel is actually created? In that case we should add the target (url String) as an argument. Maybe the decision could even be delegated to the channel factory itself, at the time that it creates the channel? Then the ClientInterceptorFilter doesn't need to know about the channel factory.


After looking into the idea of moving the decision to filter into the channel factory, I think it is better off leaving it as-is because it creates a bit of ripple.

1. The decision is still not made until the channel created as the call chain is 
```
ChannelFactory#createChannel
    -> ClientInterceptorsConfigurer.configureInterceptors
        -> ClientInterceptorFilter.shouldInclude
```
2. The createChannel method leaves "all things interceptors" in the ClientInterceptorsConfigure (seems a bit more encapsulated).

spring-grpc-core/src/test/java/org/springframework/grpc/client/ClientInterceptorsConfigurerTests.java
```diff
@@ -168,6 +169,35 @@ void whenBlendInterceptorsTrueThenGlobalInterceptorsBlended() {
...
 	}
 
+	@Nested
+	class WithInterceptorFilters {
+
+		@Test
+		void whenFilterExcludesOneGlobalInterceptor_thenBuilderGetsOnlyAllowedOnes() {
+			ManagedChannelBuilder<?> builder = Mockito.mock();
+			ClientInterceptorsConfigurerTests.this.contextRunner()
+				.withUserConfiguration(GlobalClientInterceptorsConfig.class, FilterConfig.class)
```
onobc, line 179: This is fine, but you can get rid of the extra config class and a bean directly like (either is fine, up to you, just passing info along)...

```java
...
.withBean(ClientInterceptorFilter.class, () -> (interceptor, factory) -> interceptor == GlobalClientInterceptorsConfig.GLOBAL_INTERCEPTOR_BAR)
```

spring-grpc-core/src/test/java/org/springframework/grpc/client/ClientInterceptorsConfigurerTests.java
```diff
@@ -168,6 +169,35 @@ void whenBlendInterceptorsTrueThenGlobalInterceptorsBlended() {
...
+				.run(context -> {
+					var factory = Mockito.mock(GrpcChannelFactory.class);
```
onobc, line 181: **[InfoSharing/Tip]** The `Mockito.mock()` API is nice because it simplfies the type. In this case you can replace 
```java
var factory = Mockito.mock(GrpcChannelFactory.class);
```
w/
```java
GrpcChannelFactory factory = Mockito.mock();
```.

Again, both are correct - just a different API that comes in handy sometimes.

**Review by onobc — CHANGES_REQUESTED** [review]
Thanks for the contribution - nice work! Also, the detailed PR description and your self-review is super helpful for your reviewers (PR reviewer empathy - thank you very much).

I have made some comments / suggestions. PTAL.

File: spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorFilter.java [hunk 3]
```diff
@@ -0,0 +1,38 @@
...
+public interface ClientInterceptorFilter {
+
+	/**
+	 * Determine whether the given {@link ClientInterceptor} should be included for the
+	 * provided {@link GrpcChannelFactory}.
+	 * @param interceptor the client interceptor under consideration
+	 * @param channelFactory the channel factory in use
+	 * @return {@code true} if the interceptor should be included; {@code false} otherwise
+	 */
+	boolean shouldInclude(ClientInterceptor interceptor, GrpcChannelFactory channelFactory);
```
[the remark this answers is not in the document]
therepanic: I think the contract should be inclusive. Let's take the same `Stream#filter` from the Java stream API as an example. It is inclusive. This applies to explicit filtering, but there are other Java implementations that do not explicitly implement filtering but fulfill the inclusive contract.

I think the example with `Stream#filter` is not bad; in essence, it fulfills the same contract as we do. We could consider renaming `shouldInclude` to the more straightforward `filter`, which I think is also not bad, and leave it inclusive.

As for making the default implementation return `true`, I honestly don't know. I'm probably more in favor of not doing it, since I marked the interface as `@FunctionalInterface` and there are no problems with returning true with a lambda, so the interface will be “clean.” This way, we will force the user to write the logic for filtering. I also assume that if we do the default implementation, it should be mentioned in the javadoc, , and this may complicate the method.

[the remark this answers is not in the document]
onobc: > I think the contract should be inclusive. Let's take the same Stream#filter from the Java stream API as an example. It is inclusive. This applies to explicit filtering, but there are other Java implementations that do not explicitly implement filtering but fulfill the inclusive contract.
> I think the example with Stream#filter is not bad; in essence, it fulfills the same contract as we do. We could consider renaming shouldInclude to the more straightforward filter, which I think is also not bad, and leave it inclusive.

👍🏻 Let's go w/ inclusive and name the method `filter`.

> As for making the default implementation return true, I honestly don't know. I'm probably more in favor of not doing it, since I marked the interface as @FunctionalInterface and there are no problems with returning true with a lambda, so the interface will be “clean.” This way, we will force the user to write the logic for filtering. I also assume that if we do the default implementation, it should be mentioned in the javadoc, , and this may complicate the method.

I think having the default method would not solve what I was aiming for anyways. Basically, I was going for a default  that "all interceptors are applied unless explicitly filtered out". It turns out that we get this behavior by simply by not defining any configured filter. I agree to not have the default method.

spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorsConfigurer.java
```diff
@@ -51,10 +54,29 @@ public ClientInterceptorsConfigurer(ApplicationContext applicationContext) {
...
+			boolean mergeWithGlobalInterceptors, GrpcChannelFactory factory) {
 		// Add global interceptors first
 		List<ClientInterceptor> allInterceptors = new ArrayList<>(this.globalInterceptors);
 		// Add specific interceptors
 		allInterceptors.addAll(interceptors);
+		// Filter all interceptors
+		if (factory != null) {
+			allInterceptors.removeIf(interceptor -> this.interceptorFilters.stream()
+				.anyMatch(filter -> !filter.shouldInclude(interceptor, factory)));
+		}
```
onobc, line 79: Consider this scenario of filter beans configured in order:

```
Filter1: interceptor != Foo
Filter2: interceptor == Foo
```
Given an input interceptor of Foo, the current impl would not include Foo.

I could see a case where an app may define a default global filter bean and then users may override that with more specific filter beans (via `@Order`).

therepanic:
I believe I have taken all your advice into account. However, I decided to leave `var factory = Mockito.mock(GrpcChannelFactory.class)` as it is. It's just that all these tests use `var`, and I think it's better to leave it that way. What do you think?

I also removed the overload. In addition, in `ClientInterceptorsConfigurer#findInterceptorFilters`, we now return a wrapper in ArrayList so that we can then do `ApplicationContextBeanLookupUtils#sortBeansIncludingOrderAnnotation(this.applicationContext,
                ClientInterceptorFilter.class, this.interceptorFilters)`. This is necessary for sorting by `@Order`.

I also reworked the logic of `ClientInterceptorFilter#filter` when we have many such beans registered. That is, if any filter returns false, we remove the interceptor. Now we need at least one filter that returns true, otherwise we remove it.

spring-grpc-core/src/main/java/org/springframework/grpc/client/ClientInterceptorsConfigurer.java
```diff
@@ -48,13 +51,19 @@ public ClientInterceptorsConfigurer(ApplicationContext applicationContext) {
...
+			boolean mergeWithGlobalInterceptors, GrpcChannelFactory factory) {
 		// Add global interceptors first
 		List<ClientInterceptor> allInterceptors = new ArrayList<>(this.globalInterceptors);
 		// Add specific interceptors
 		allInterceptors.addAll(interceptors);
+		if (!this.interceptorFilters.isEmpty()) {
+			// Filter all interceptors
+			allInterceptors.removeIf(interceptor -> this.interceptorFilters.stream()
+				.noneMatch(filter -> filter.filter(interceptor, factory)));
+		}
```
onobc, line 66: The changes all LGTM @therepanic . However, I am still not sure this solves the use case I outlined. Furthermore, I think having multiple filters is complicating things. I wonder if instead we allow only a single `ClientIntercerptorFilter` bean that is used to configure the client interceptors that should apply to the factory. This is likely an edge-case or at best a secondary usecase. 

@dsyer @therepanic  wdyt?

therepanic: Initially, I also wanted to make it possible to declare a single bean, as it is indeed simpler. However, I considered that this was “incorrect,” although in essence it probably makes no difference. In general, I do not believe that it is problematic to allow the declaration of one bean or several beans. I will push the option with one bean, and we will see how it looks. Personally, I think both the option with multiple beans and the option with a single bean look good.

therepanic:
In the catch block in `ClientInterceptorsConfigurer#findInterceptorFilter`, I return null. I could return something like `() -> true` and not check for null in `ClientInterceptorsConfigurer#configureInterceptors`, but I thought this was better because we don't create an extra object in memory. What do you think?

onobc:
Thanks for another great contribution @therepanic !

Closing in favor of https://github.com/spring-projects/spring-grpc/commit/161ad69de29b0f52b20d5373e48e333bfaa5d941

The remaining piece is to add some docs for this. I will try to beat you to it sometime today though. :)

Issue #197, opened 2025-06-18
Make GrpcServerFactoryAutoConfiguration conditional on the presence of an actual service to bind to


dsyer:
It's not a terrible idea to write a unit test for a client app that doesn't have any `BindableService` beans (and hence doesn't want to start a server on port 9090). You can explicitly stop the server from starting by adding `spring.grpc.server.enabled=false` but probably it should be automatic.

therepanic:
Hi, @dsyer. Isn't this mechanism already implemented in 2eaaa8f8? Also, a test has already been attached to `GrpcServerAutoConfiguration`:
```java
@Test
   void whenNoBindableServicesRegisteredAutoConfigurationIsSkipped() {
       new ApplicationContextRunner().withConfiguration(AutoConfigurations.of(GrpcServerAutoConfiguration.class))
           .run((context) -> assertThat(context).doesNotHaveBean(GrpcServerAutoConfiguration.class));
}
```

dsyer:
That's interesting. I have a test that is definitely starting the server and doesn't need to. I have to look in more detail at why that happens. Maybe there is a reflection service or a health service or something that shouldn't be auto-registered.

UPDATE: that was it:

```
GrpcServerAutoConfiguration matched:
      - @ConditionalOnClass found required class 'io.grpc.BindableService' (OnClassCondition)
      - @ConditionalOnProperty (spring.grpc.server.enabled) matched (OnPropertyCondition)
      - @ConditionalOnBean (types: io.grpc.BindableService; SearchStrategy: all) found beans 'serverReflection', 'grpcHealthService' (OnBeanCondition)
```

therepanic:
Yes, I also just noticed the `BindableService` registration in `GrpcServerHealthAutoConfiguration`. Apparently, it is also present in `serverReflection`.

What can be done about this? We need to somehow ignore the registration of these `BindableService` in client applications.

dsyer:
We can just not create those beans `@ConditionalOnBean(BindableService.class)`

therepanic:
Okay, if you don't mind, I can do the PR. Just one thing I'm wondering about. I think we should do a test to identify this problem, so that it doesn't happen again in the future. Where would be the best place to put it?

dsyer:
I'm sure you'll find a place (next to the test you just quoted maybe?). Isn't there already an integration test for the reflection service and health service?

PR #198, opened 2025-06-19
Register health and reflection services only if `BindableService` beans exist

therepanic:
We fixed this problem by simply adding the corresponding annotation `@ConditionalOnBean(BindableService.class)` to `GrpcServerReflectionAutoConfiguration` and `GrpcServerHealthAutoConfiguration`.

Now we have a question about how to test this mechanism. In `GrpcServerAutoConfigurationTests`, we have the following test:
```java
@Test
void whenNoBindableServicesRegisteredAutoConfigurationIsSkipped() {
    new ApplicationContextRunner().withConfiguration(AutoConfigurations.of(GrpcServerAutoConfiguration.class))
        .run((context) -> assertThat(context).doesNotHaveBean(GrpcServerAutoConfiguration.class));
}
```
We could add the appropriate auto-configurations and the test would indeed check everything we need. However, I believe this is not the correct approach. This test, as it stands, is completely correct and checks the mechanism of not creating `GrpcServerAutoConfiguration`. Also, if we add these auto-configurations, we are not immune to problems in the future where we add another auto-configuration that will create `BindableService` and we forget to specify it here. Then we will see the same problem again.

I thought about it and came up with what I think is the best option. We could add an integration test that loads the client application, along with all the auto-configurations for the client, and checks that no `BindableService` is registered. Before this fix, there were two. So the test is working.

Fix: #197

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfigurationTests.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfigurationTests.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.springframework.grpc.autoconfigure.server;
 
 import io.grpc.BindableService;
@@ -9,12 +24,19 @@ import org.springframework.grpc.server.lifecycle.GrpcServerLifecycle;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+/**
+ * Tests for {@link GrpcServerReflectionAutoConfiguration}.
+ *
+ * @author Haris Zujo
+ * @author Chris Bono
+ * @author Andrey Litvitski
+ */
 public class GrpcServerReflectionAutoConfigurationTests {
 
 	private ApplicationContextRunner contextRunner() {
 		return new ApplicationContextRunner()
 			.withConfiguration(AutoConfigurations.of(GrpcServerReflectionAutoConfiguration.class))
-			.withBean("noopServerLifecylcle", GrpcServerLifecycle.class, Mockito::mock)
+			.withBean("noopServerLifecycle", GrpcServerLifecycle.class, Mockito::mock)
 			.withBean(BindableService.class, Mockito::mock);
 	}
 
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfiguration.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.springframework.grpc.autoconfigure.server;
 
 import org.springframework.boot.autoconfigure.AutoConfiguration;


therepanic:
Hmm, I didn't notice that the tests for these configurations fail. They are all pretty much the same and check something like this:
```java
.run((context) -> assertThat(context).hasSingleBean(BindableService.class));
```
This worked before because we loaded a configuration that created the corresponding `BindableService` without checking.

To solve this problem, I suggest mock a separate `BindableService` bean and checking that there are 2 `BindableService` beans in the context.

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/health/GrpcServerHealthAutoConfigurationTests.java
```diff
@@ -55,7 +55,8 @@ class GrpcServerHealthAutoConfigurationTests {
 
 	private ApplicationContextRunner contextRunner() {
 		return new ApplicationContextRunner()
-			.withConfiguration(AutoConfigurations.of(GrpcServerHealthAutoConfiguration.class));
```
onobc, line 58: I think the proper test to add here should answer the following: 

"The auto-config now relies on a bindable service bean to be defined or the auto-config is skipped". This is the simplest test to do that IMO:

```java
@Test
void whenNoBindableServiceDefinedDoesNotAutoConfigureBean() {
new ApplicationContextRunner()
		.withConfiguration(AutoConfigurations.of(GrpcServerHealthAutoConfiguration.class))
		.run((context) -> assertThat(context).doesNotHaveBean(GrpcServerHealthAutoConfiguration.class));
}


```

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfigurationTests.java
```diff
@@ -14,45 +16,46 @@ public class GrpcServerReflectionAutoConfigurationTests {
...
-		this.contextRunner().run((context) -> assertThat(context).hasSingleBean(BindableService.class));
+		this.contextRunner().run((context) -> assertThat(context).getBeanNames(BindableService.class).hasSize(2));
 	}
 
 	@Test
 	void whenServerEnabledPropertySetTrueThenAutoConfigurationIsNotSkipped() {
 		this.contextRunner()
 			.withPropertyValues("spring.grpc.server.enabled=true")
-			.run((context) -> assertThat(context).hasSingleBean(BindableService.class));
+			.run((context) -> assertThat(context).getBeanNames(BindableService.class).hasSize(2));
```
onobc, line 58: I think you will also want to add another test here as described in https://github.com/spring-projects/spring-grpc/pull/198/files#r2157685154

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfigurationTests.java
```diff
@@ -14,45 +16,46 @@ public class GrpcServerReflectionAutoConfigurationTests {
...
 			.withConfiguration(AutoConfigurations.of(GrpcServerReflectionAutoConfiguration.class))
-			.withBean("noopServerLifecylcle", GrpcServerLifecycle.class, Mockito::mock);
+			.withBean("noopServerLifecylcle", GrpcServerLifecycle.class, Mockito::mock)
+			.withBean(BindableService.class, Mockito::mock);
 	}
 
 	@Test
 	void whenReflectionEnabledFlagNotPresentThenCreateDefaultBean() {
-		this.contextRunner().run((context -> assertThat(context).hasSingleBean(BindableService.class)));
+		this.contextRunner().run((context -> assertThat(context).getBeanNames(BindableService.class).hasSize(2)));
```
onobc, line 25: I see what you are doing here, verifying that our bean we used to turn on the auto-config and then the one that the auto-config creates. I don't think it is important to count the beans but rather we only want to test what the auto-config produces when its "on" or "off".  In this case, it should create a BindableService w/ name "serverReflection". AssertJ provides a clean way to do this like:
```java
@Test
void whenReflectionEnabledFlagNotPresentThenCreateDefaultBean() {
        this.contextRunner().run((context ->
				assertThat(context).hasBean("serverReflection")));
}
```

Wdyt?

therepanic: Yes, I completely agree, that would indeed be better, thanks

samples/grpc-server/src/test/java/org/springframework/grpc/sample/GrpcClientApplicationTests.java
```diff
@@ -21,6 +22,20 @@
...
+	@SpringBootTest(properties = "service.enabled=false")
+	class NoBindableServicesWhenDisabled {
+
+		@Autowired
+		private ApplicationContext context;
+
+		@Test
+		void noBindableServicesAreCreated() {
+			assertThat(context.getBeanNamesForType(BindableService.class)).isEmpty();
+		}
```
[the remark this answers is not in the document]
onobc: > Also, if we add these auto-configurations, we are not immune to problems in the future where we add another auto-configuration that will create BindableService and we forget to specify it here

In general, we add a thorough auto-configuration tests for each new auto-configuration added to the system. It is interesting because it is not quite a unit test (as we use the ApplicationContextRunner which creates a context) and not quite a full integration test. I think of "auto-configuration tests" as "lightweight integration tests focused only on the aspect of auto-configuration conditions and results".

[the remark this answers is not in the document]
therepanic: The main idea behind this test was to cover cases where we added auto-configuration that registers the `BindableService` bean and did not specify `@ConditionalOnProperty`. In this case, the test would fail. However, as you described in https://github.com/spring-projects/spring-grpc/pull/198#discussion_r2157685154, it is indeed sufficient for us to add the test you specified and add it in the future if there are auto-configurations of this kind.

onobc (changes requested):
Great work @therepanic ! I appreciate the deep thought and the back and forth we have on these reviews - this is why we do them.

I have a couple of comments/questions to address but other than that LGTM. Oh, and thank you again!

therepanic:
I did it a little differently. Instead of using specific bean names, I simply did:
```java
.hasSingleBean(GrpcServerReflectionAutoConfiguration.class)
```

You did the same in `GrpcServerHealthAutoConfigurationTests`.

I also added a separate polish commit, added a license for `GrpcServerReflectionAutoConfiguration` that we worked with, and for class test.

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerReflectionAutoConfigurationTests.java
```diff
@@ -9,50 +24,67 @@
...
+		this.contextRunner()
+			.run((context -> assertThat(context).hasSingleBean(GrpcServerReflectionAutoConfiguration.class)));
 	}
 
 	@Test
 	void whenReflectionEnabledThenCreateBean() {
 		this.contextRunner()
 			.withPropertyValues("spring.grpc.server.reflection.enabled=true")
-			.run((context) -> assertThat(context).hasSingleBean(BindableService.class));
+			.run((context) -> assertThat(context).hasSingleBean(GrpcServerReflectionAutoConfiguration.class));
```
onobc, line 60: **[TL;DR]** Using the presence of the autoconfig bean rather than the bindable service is the correct approach (good catch - more below).

So for auto-configuration tests I typically attack 2 vectors.

**First:** The 1st is all of the conditions that enable/disable the auto-configuration. In this example:

```
@AutoConfiguration(before = GrpcServerFactoryAutoConfiguration.class)
@ConditionalOnGrpcServerEnabled
@ConditionalOnClass(ProtoReflectionServiceV1.class)
@ConditionalOnProperty(name = "spring.grpc.server.reflection.enabled", havingValue = "true", matchIfMissing = true)
@ConditionalOnBean(BindableService.class)
public class GrpcServerReflectionAutoConfiguration {
```
And for these tests it is sufficient to just check that the auto-config bean is or is not in the context based on the test conditions (this is the change you made here 👍🏻 ).

**Second:** The 2nd is to verify that the beans the auto-config is supposed to configure, are in fact configured. In this example it would be:

```
@Bean
public BindableService serverReflection() {
	return ProtoReflectionServiceV1.newInstance();
}
```
And the test is is as simple as:
```
this.contextRunner().run((context ->
				assertThat(context).hasBean("serverReflection")));
```

> [!NOTE]
> I do realize that the way that I wrote this test initially was using the `BindableService` created by the auto-config in order to "verify" behavior (vs. the autoconfiguration bean). I do think I should have just checked if the auto-config bean is present. Then only in 1 test we could add that would check if the bean is actually created.

onobc (changes requested):
Thanks for another great contribution @therepanic !

This one is just about wrapped up. I have one minor comment to address and the other comments are just info sharing etc.. PTAL.

> [!NOTE]
> For future PRs, I do have one request though, not to squash/force push over the original commit. This helps w/ reviewer empathy as it is easier to do subsequent rounds of reviews if we leave the original commit as-is (vs. squash and force push). Not a huge deal but it let's me see easily what has changed since I reviewed it last. We will of course rebase and sqaush before we merge into main.

therepanic:
Thank you, I'll keep that in mind about squashing. This time, I'll squash everything again and force push, as I believe this is the final change, but in the future, I'll make separate commits after review rounds.

onobc:
Thanks again for another contribution @therepanic ! Closing in favor of https://github.com/spring-projects/spring-grpc/commit/fdb7129ada05f13ba2e27b3ac9fe5a4247a0fadb. Also, you can see where I added a [minor polish commit](https://github.com/spring-projects/spring-grpc/commit/be6318febfe0f82a3bf15985bf8ea414a73126da) to rename some of the test method names for consistency (really unrelated to your new changes).

PR #200, opened 2025-06-23
Add docs client interceptor filter

onobc:
Adds section on how to use the `ClientInterceptorFIlter`.

Screenshot 2025-06-23 at 15 08 47

@therepanic PTAL


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
```diff
@@ -150,12 +150,11 @@ public Health getHealth() {
 			return this.health;
 		}
 
-		private final Map<String, ?> serviceConfig = new HashMap<>();
-
 		/**
-		 * The service config to use for the channel.
-		 * @return the service config
+		 * Map representation of the service config to use for the channel
```
onobc, line 154: The config props generator was unable to pickup the docs from the getter - moved it to the field instead.

therepanic:
I think it's described very well

onobc:
Thank you for the review @therepanic

PR #201, opened 2025-06-23
Add docs for grpc-services dependency

### LINKED ISSUE #189: Consider removing optional from grpc-services or update the docs to raise more awareness
Hi team,

the issue is that although the sample rely explicitly in the [grpc-services](https://github.com/spring-projects/spring-grpc/blob/main/samples/grpc-server/build.gradle#L33) , the out-of-the-box experience could be better if either the documentation clearly states the need for it when using [GrpcHealth](https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-docs/src/main/antora/modules/ROOT/pages/health.adoc) or simply that similar to the [client-starter that relies transitively in grpc-stub](https://mvnrepository.com/artifact/org.springframework.grpc/spring-grpc-client-spring-boot-starter/0.8.0), the [server-starter could rely transitively on grpc-services instead of making it optional](https://mvnrepository.com/artifact/org.springframework.grpc/spring-grpc-server-spring-boot-starter/0.8.0).

And then if anyone would not need it, could do the exclusion when declaring the dependency. That is because I would _assume_ that anyone using the server-starter would expect that ootb it would enable GrpcHealth as well.

But if for any given reason the optional must stay, at least the docs could be more explicitly to avoid investing time chasing the reason why the GrpcHealth was not available when starting a server.

Thank you in advance.

onobc:
Clarifies in the docs that the grpc-services dependency is needed for health and reflection services to be autoconfigured.

Resolves #189 

I also added a section for the reflection service to clarify the same. @lobaorn-bitso PTAL 

Screenshot 2025-06-23 at 17 19 55

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/pages/health.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/pages/health.adoc
@@ -5,6 +5,8 @@
 Spring gRPC autoconfigures the standard https://grpc.io/docs/guides/health-checking/[gRPC Health service] for performing health check calls against gRPC servers.
 The health service is registered with the gRPC server and a `HealthStatusManager` bean is provided that can be used to update the health status of your services.
 
+IMPORTANT: The health service resides in the `io.grpc:grpc-services` library which is marked as `optional` by Spring gRPC. You must add this dependency to your application in order for it to be autoconfigured.
+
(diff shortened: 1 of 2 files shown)


lobaorn-bitso (approved):
Thanks a lot @onobc , it is much clearer now. FYI @chochos

Issue #203, opened 2025-06-25
Compatible issue with old grpc version


GAATTC0:
**Environment**

I use grpc `io.grpc:protoc-gen-grpc-java:1.15.1` for code generation:

```
protobuf {
    plugins {
        id("grpc") {
        artifact = "io.grpc:protoc-gen-grpc-java:1.15.1"
        }
    }
}
```
**Cause**

the generated BlockingStub extends `AbstractStub` by this version, not `AbstractBlockingStub`:

![Image](https://github.com/user-attachments/assets/202550af-da9e-427d-b155-9f23d8d93c13)

when creating stub on start, it is not match correctly in BlockingStubFactory:

![Image](https://github.com/user-attachments/assets/8ca33116-250c-4317-ac41-c24b9ec602d6)

because it's just judge if `extends AbstractBlockingStub` and `!type.getSimpleName().contains("BlockingV2")`

i tried change the plugin version to 1.30.0, BlockingStub in generated code extends AbstractBlockingStub, it works now.

**Issue**

i think we can change the predicate to make compatible with old version:

```
    public class BlockingStubFactory extends AbstractStubFactory<AbstractBlockingStub<?>> implements Ordered {

        @Override
        public boolean supports(Class<?> type) {
            return (AbstractStubFactory.supports(AbstractBlockingStub.class, type) && !type.getSimpleName().contains("BlockingV2"))
                    || type.getSimpleName().endsWith("BlockingStub");
        }
    }
```

dsyer:
I don't see much value in supporting older versions for this when we don't really support it generally - I mean, we don't want to be chasing every single tiny incompatibility in older versions, when users can just upgrade (always a better option). On the other hand this is a minor change, so if it makes one person happy, I guess I'm OK with that too.

GAATTC0:
> I don't see much value in supporting older versions for this when we don't really support it generally - I mean, we don't want to be chasing every single tiny incompatibility in older versions, when users can just upgrade (always a better option). On the other hand this is a minor change, so if it makes one person happy, I guess I'm OK with that too.

Thanks for reply! 
there's a case in my team: we put all proto files in a independent repo which manage grpc version self, and other serveral services depends it, there are different grpc versions around them because of they maintenanced by different team.
The best way is to upgrade and unify version, but difficult in reality, and grpc's api is compatible along the verions;
So make spring-grpc compatible with old version is a easy & harmless way to fix the case above, i wish this change will help more people facing the problem like mine

PR #205, opened 2025-06-25
compatible for old grpc version

GAATTC0:
#203


dsyer:
The build failures tell you how to fix it.

PR #206, opened 2025-06-27
Fix in-process channel target doc

Meijuh:
I finally found some time to test the changes in https://github.com/spring-projects/spring-grpc/pull/187 and check the docs from https://github.com/spring-projects/spring-grpc/pull/194. The implementation works, however this PR shows there is a mismatch between the docs at https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-docs/src/main/antora/modules/ROOT/pages/server.adoc#inprocess-server and the implementation at https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-core/src/main/java/org/springframework/grpc/client/InProcessGrpcChannelFactory.java#L54: heading `//` should not be part of the target. I am assuming the docs are wrong, as it seems this existing approach is followed: https://yidongnan.github.io/grpc-spring-boot-starter/en/server/configuration.html.

I have a question about the chosen approach though. By specifying `in-process` as a scheme, it seems as if this is a valid name resolver compliant scheme, but I could not find it at https://grpc.io/docs/guides/custom-name-resolution, nor anywhere else as far as I can see. Also I could not find a custom `NameResolverProvider` (https://grpc.github.io/grpc-java/javadoc/io/grpc/NameResolverProvider.html) that resolves the `in-process` scheme. This means `in-process` only exists somewhat as a scheme in `spring-grpc` but not in the gRPC ecosystem. This is just an observation and might confuse other devs; it could be the right approach however.

As an aside; this PR also includes a fix for a typo in the docs of methods `supports(String target)`.

Another question that I have; with the added support for in-process servers; is it possible to exclude binding the "Health" service? I would never perform health checks on in-process servers.

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
@@ -72,7 +72,7 @@ public class DefaultGrpcChannelFactory<T extends ManagedChannelBuilder<T>>
 
 	/**
 	 * Whether this factory supports the given target string. The target can be either a
-	 * valid nameresolver-compliant URI, an authority string as described in
+	 * valid nameresolver-compliant URI or an authority string as described in
 	 * {@link Grpc#newChannelBuilder(String, ChannelCredentials)}.
 	 * @param target the target string as described in method javadocs
 	 * @return true unless the target begins with 'in-process:'
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelFactory.java
@@ -31,7 +31,7 @@ public interface GrpcChannelFactory {
 
 	/**
 	 * Whether this factory supports the given target string. The target can be either a
-	 * valid nameresolver-compliant URI, an authority string as described in
+	 * valid nameresolver-compliant URI or an authority string as described in
 	 * {@link Grpc#newChannelBuilder(String, ChannelCredentials)}.
 	 * @param target the target string as described in method javadocs
 	 * @return whether this factory supports the given target string
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/InProcessGrpcChannelFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/InProcessGrpcChannelFactory.java
@@ -44,7 +44,7 @@ public class InProcessGrpcChannelFactory extends DefaultGrpcChannelFactory<InPro
 
 	/**
 	 * Whether this factory supports the given target string. The target can be either a
-	 * valid nameresolver-compliant URI, an authority string as described in
+	 * valid nameresolver-compliant URI or an authority string as described in
 	 * {@link Grpc#newChannelBuilder(String, ChannelCredentials)}.
 	 * @param target the target string as described in method javadocs
 	 * @return true if the target begins with 'in-process:'


onobc:
Hi @Meijuh 

Thanks for taking the time to review the changes and give this feedback. I have grouped my replies by "topic/area".

### Impl concerns / doc mismatch 

I believe the disconnect here is that the `supports` method is only called by the [CompositeGprcChannelFactory](https://github.com/spring-projects/spring-grpc/blob/cf15279c1b79bfeffe4d4894d4fb073facbf1489/spring-grpc-core/src/main/java/org/springframework/grpc/client/CompositeGrpcChannelFactory.java#L55) and then the base [DefaultGrpcChannelFactory](https://github.com/spring-projects/spring-grpc/blob/cf15279c1b79bfeffe4d4894d4fb073facbf1489/spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java#L96) calls into the factory [newChannelBuilder](https://github.com/spring-projects/spring-grpc/blob/cf15279c1b79bfeffe4d4894d4fb073facbf1489/spring-grpc-core/src/main/java/org/springframework/grpc/client/InProcessGrpcChannelFactory.java#L59) which then in this case strips the `in-process:` prefix from the target prior to passing into the `InProcessChannelBuilder.forName`.

> This means in-process only exists somewhat as a scheme in spring-grpc but not in the gRPC ecosystem. This is just an observation and might confuse other devs; it could be the right approach however.

Right, this is not a scheme known to `gRPC` core nor `gRPC Java` but only to `gRPC ecosystem` and `Spring gRPC`.

> [!NOTE]
> Here is the legend of names to repo as they are in my brain:
> - `gRPC` core -> https://github.com/grpc/grpc
> - `gRPC Java` -> https://github.com/grpc/grpc-java
> - `gRPC Ecosystem` -> https://github.com/grpc-ecosystem/grpc-spring
> - `Spring gRPC` -> https://github.com/spring-projects/spring-grpc
> 
> The name `gRPC Ecosystem` for the previous starter repo is a very broad name. Me and you (based on your description) think of the "ecosystem" as the core plus all of its implementations, etc.. But we are on the same page.

We think this is a fair approach because:
1. the `gRPC Ecosystem` has heavy usage and been around a while (i.e. precedence has been set / user's are probably used to using this `in-process:`) and 
2. the other requirements we were after (composite switching to other factories), this approach "checked the most boxes" during implementation. 

PTAL and let us know if this eases your concerns. Either way, these changes are correct and this PR is needed. I will merge once we close our discussion though.

### Filtering out services

> Another question that I have; with the added support for in-process servers; is it possible to exclude binding the "Health" service? I would never perform health checks on in-process servers.

Ahh, great point/catch. Looks like this is not currently possible. 

We did solve this for the client interceptors by introducing a [ClientInterceptorFilter](https://docs.spring.io/spring-grpc/reference/client.html#global-client-interceptor-filtering). However, on the server-side the only hook we have is `ServerBuilderCustomizer` but by the time they are [applied](https://github.com/spring-projects/spring-grpc/blob/cf15279c1b79bfeffe4d4894d4fb073facbf1489/spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java#L147) the services have already been [added to the builder](https://github.com/spring-projects/spring-grpc/blob/cf15279c1b79bfeffe4d4894d4fb073facbf1489/spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java#L164). 

I have created [#207](https://github.com/spring-projects/spring-grpc/issues/207) and marked it with milestone `0.9.0` which we are releasing soon. If you would like to contribute this improvement it would be most welcomed. However, if you don't have time to implement this soon just lmk asap so I can see if I can squeeze it in on my side. 

Thanks again.

-chris

onobc:
@Meijuh - looks like our friend @therepanic is interested in implementing #207 !

onobc:
@Meijuh do you mind signing your commit so the [DCO will pass](https://github.com/spring-projects/spring-grpc/pull/206/checks?check_run_id=44943322320) the checks?

Meijuh:
@onobc Friday I already did a force push to include signoffs to my commits, but it seems the DCO check does not pick up the force push (or restart the check). Here https://github.com/Meijuh/spring-grpc/commits/main/ you can see my top 2 commits have been signed off. Any idea what I can do about this?

About using `in-process` as a scheme; if this is what users are expecting, then this is perfectly okay.

onobc:
> @onobc Friday I already did a force push to include signoffs to my commits, but it seems the DCO check does not pick up the force push (or restart the check). Here https://github.com/Meijuh/spring-grpc/commits/main/ you can see my top 2 commits have been signed off. Any idea what I can do about this?
> 
> About using `in-process` as a scheme; if this is what users are expecting, then this is perfectly okay.

Hi @Meijuh , I think you may need to rebase as well for it to pickup the changes. Take a peek at the "details" of the DCO failure and it gives some info on the rebase. If that does not work then I will dig in further. Thanks

Meijuh:
@onobc it should be good now, there was a mismatch between the email I used for signing off, and committing.

onobc:
Thanks for the contribution @Meijuh - much appreciated. Closing in favor of https://github.com/spring-projects/spring-grpc/commit/378c1d08e63f6456bd6dadc26fc99b11a0b2380b. I also added a follow on commit as I noticed after the review that we also needed to update the client.adoc [here](https://github.com/spring-projects/spring-grpc/commit/05f07582bf1cff0501bf91c8c7f4864f5b42ae5a).

PR #209, opened 2025-06-29
Add support for server interceptor filtering

### LINKED ISSUE #208: Add ability to filter out server interceptors from server factories
Similar to the need in https://github.com/spring-projects/spring-grpc/issues/195 - we should allow a user-configured bean that can filter which server global interceptors are included/excluded from specific server factories.

onobc:
> [!WARNING]
> **WIP**
> This is being reworked once #212  is merged

This commit adds support for filtering the server-side global interceptors.

The server factory is now passed down through the service discoverer and configurer to allow the filter to take into account the server factory in use.

The service discoverer contract has also been minimized to remove the unused `findService(String)` method.

Resolves #208

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/server/service/DefaultGrpcServiceConfigurer.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/server/service/DefaultGrpcServiceConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/server/ServletGrpcServerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.grpc.server;
+
+import io.grpc.Server;
+import io.grpc.ServerServiceDefinition;
+
+/**
+ * Marker interface for {@link GrpcServerFactory} that is to be handled by the servlet
```
onobc, line 23: Not super excited about this but the alternative is to pass in null in this case but then the filter implementations must take this into account and check for null and then assume null means servlet.

PR #211, opened 2025-06-30
Fix serviceConfig within scope of copy

### LINKED ISSUE #210: serviceConfig should be involved in the copy() of GrpcClientProperties
The newly added serviceConfig field in the ChannelConfig class was not copied in the copy() method, which causes the copied object to lose the service configuration.

NeatGuyCoding:
Fix #210

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
@@ -153,9 +153,9 @@ public class GrpcClientProperties implements EnvironmentAware, VirtualTargets {
 		/**
 		 * Map representation of the service config to use for the channel
 		 */
-		private final Map<String, ?> serviceConfig = new HashMap<>();
+		private final Map<String, Object> serviceConfig = new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


onobc:
Hi @NeatGuyCoding , good catch and thank you for the contribution. Can you please follow [these instructions](https://github.com/spring-projects/spring-grpc/pull/211/checks?check_run_id=45031342353) and sign the commit and we will get it merged after that? Thanks

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcClientProperties.java
```diff
@@ -349,6 +349,8 @@ public ChannelConfig copy() {
 			copy.defaultDeadline = this.defaultDeadline;
 			copy.health.copyValuesFrom(this.getHealth());
 			copy.ssl.copyValuesFrom(this.getSsl());
+			copy.serviceConfig.clear();
```
onobc, line 352: I think we can drop this `clear` call.

NeatGuyCoding: ok, I will fix this

NeatGuyCoding:
all fixed, thanks a lot @onobc

PR #212, opened 2025-06-30
Split service discoverer and configurer

onobc:
This commit simplifies the DefaultGrpcServiceDiscoverer by removing its dependence on the DefaultGrpcServiceConfigurer. Now the expectation is the discoverer returns a spec of the service definition rather than a bound service. The configurer then accepts that spec and binds and configures it.

See #208

### DIFF
--- a/spring-grpc-core/src/test/java/org/springframework/grpc/server/service/DefaultGrpcServiceDiscovererTests.java
+++ b/spring-grpc-core/src/test/java/org/springframework/grpc/server/service/DefaultGrpcServiceDiscovererTests.java
@@ -17,11 +17,10 @@
 package org.springframework.grpc.server.service;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.springframework.grpc.server.service.DefaultGrpcServiceDiscovererTests.DefaultGrpcServiceDiscovererTestsConfig.SERVICE_A;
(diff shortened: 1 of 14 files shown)


spring-grpc-core/src/test/java/org/springframework/grpc/server/service/DefaultGrpcServiceDiscovererTests.java
```diff
@@ -43,28 +43,55 @@
...
-							DefaultGrpcServiceDiscovererTestsConfig.SERVICE_DEF_A);
-				TestServiceConfigurer configurer = context.getBean(TestServiceConfigurer.class);
-				assertThat(configurer.invocations).hasSize(2);
-				assertThat(configurer.invocations.keySet()).containsExactly(SERVICE_B, SERVICE_A);
-				assertThat(configurer.invocations).containsEntry(SERVICE_B, null);
-				assertThat(configurer.invocations).hasEntrySatisfying(SERVICE_A, (serviceInfo) -> {
-					assertThat(serviceInfo.interceptors()).isEmpty();
-					assertThat(serviceInfo.interceptorNames()).isEmpty();
-					assertThat(serviceInfo.blendWithGlobalInterceptors()).isFalse();
-				});
```
onobc, line 62: This is one of the fruits of this simplification (goodbye verifying cofigurer things in discoverer tests)

PR #213, opened 2025-06-30
Add ability to filter out services from being bound to server factories

therepanic:
With this change, we will give users the ability to configure their filters for the server-side factory.

I created an interface similar to the ones we already have, and in `DefaultGrpcServerFactory` we use it to add/not add a service. Overall, nothing special, but I will leave a couple of comments on the points that bother me in some way.


spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
```diff
@@ -68,17 +68,21 @@ public class DefaultGrpcServerFactory<T extends ServerBuilder<T>> implements Grp
...
 	public DefaultGrpcServerFactory(String address, List<ServerBuilderCustomizer<T>> serverBuilderCustomizers,
-			KeyManagerFactory keyManager, TrustManagerFactory trustManager, ClientAuth clientAuth) {
+			KeyManagerFactory keyManager, TrustManagerFactory trustManager, ClientAuth clientAuth,
+			ServerServiceDefinitionFilter serviceFilter) {
 		this(address, serverBuilderCustomizers);
 		this.keyManager = keyManager;
 		this.trustManager = trustManager;
 		this.clientAuth = clientAuth;
+		this.serviceFilter = serviceFilter;
 	}
```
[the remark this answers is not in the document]
dsyer: I don't like the way that some of the new constructors use `Optional` and this one doesn't (not optional but `@Nullable` seems better in a mostly internal API). I'm neutral on whether to add another constructor (but mostly against).

spring-grpc-core/src/main/java/org/springframework/grpc/server/ServerServiceDefinitionFilter.java
```diff
@@ -0,0 +1,38 @@
...
+public interface ServerServiceDefinitionFilter {
+
+	/**
+	 * Determine whether the given {@link ServerServiceDefinition} should be included for
+	 * the provided {@link GrpcServerFactory server factory}.
+	 * @param serviceDefinition the gRPC service definition under consideration.
+	 * @param serverFactory the server factory in use.
+	 * @return {@code true} if the service should be included; {@code false} otherwise.
+	 */
+	boolean filter(ServerServiceDefinition serviceDefinition, GrpcServerFactory serverFactory);
```
dsyer, line 36: This seems wrong - there are no methods on `GrpcServerFactory` that we would expect to be useful to implementations of this interface, so it's confusing to pass it in. I don't know if removing it will cause more trouble than it is worth but I'd like to at least see that option.

therepanic:
Thanks for the review. I removed the second argument from `ServerServiceDefinitionFilter` and also added the `@Nullable` annotation to its class constructor. Instead of using `Option` in autoconfigure, I also use the `@Nullable` annotation.

spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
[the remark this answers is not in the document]
onobc: Yeh, we typically rarely use `Optional` for things besides returns types. For consistency I would say lets go w/ null and `@Nullable` as well.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerFactoryConfigurations.java
```diff
@@ -75,7 +78,7 @@ ShadedNettyGrpcServerFactory shadedNettyGrpcServerFactory(GrpcServerProperties p
 						: io.grpc.netty.shaded.io.netty.handler.ssl.util.InsecureTrustManagerFactory.INSTANCE;
 			}
 			ShadedNettyGrpcServerFactory factory = new ShadedNettyGrpcServerFactory(properties.getAddress(),
-					builderCustomizers, keyManager, trustManager, properties.getSsl().getClientAuth());
+					builderCustomizers, keyManager, trustManager, properties.getSsl().getClientAuth(), serviceFilter);
```
onobc, line 81: We will need to do the same thing for the `InProcessGrpcServerFactory` and `TestInProcessGrpcServerFactory`.

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfigurationTests.java
```diff
@@ -407,6 +411,47 @@ void nettyServerFactoryAutoConfiguredWithSsl() {
 				NettyGrpcServerFactory.class, "myhost:6160", "nettyGrpcServerLifecycle");
 	}
 
+	@Nested
+	class WithFilter {
```
onobc, line 415: Nice test!

spring-grpc-spring-boot-autoconfigure/src/test/java/org/springframework/grpc/autoconfigure/server/GrpcServerAutoConfigurationTests.java
```diff
@@ -66,6 +69,7 @@
  * Tests for {@link GrpcServerAutoConfiguration}.
  *
  * @author Chris Bono
+ * @author Andrey Litvitski
```
onobc, line 72: Please add yourself as author on the other updated files (factories namely)

onobc (changes requested):
Thanks @therepanic - looking good. A couple of comments to address and then we will be good to go.

onobc:
Closing in favor of #216

PR #218, opened 2025-07-02
Gh 207 208 explore filter options

onobc:
This is the 2 commits in my PR for server interceptor filters on top of main.

Then I explore moving the filters as follows:

### Server Interceptor Filter

- move filter instance from configuruer to InProcessGrpcServerFactory
- add `supports(interceptor, service)` on GrpcServerFactory w/ default return `true`
- modify configurer to ask the factory being configured if it supports the interceptor/service
- override `supports` in InProcessGrpcServerFactory to ask its filter instance if it supports it

**Tradeoffs**
✅  This removes the server factory from the interceptor filter contract

❌   Currently the ServerInterceptorFilter is only used by InProcessGrpcServerFactory so it may be misleading by its name for users thinking it may be used by all servers

### Service Filter

- move filter instance from DefaultGrpcServerFactory to InProcessGrpcServerFactory
- remove use of filter in `addService` of DefaultGrpcServerFactory
- override `addService` in InProcessGrpcServerFactory to ask its filter instance if it should add it

**Tradeoffs**
✅  This removes the server factory from the interceptor filter contract

❌ Currently the ServerServiceDefinitionFilter is only used by InProcessGrpcServerFactory so it may be misleading by its name for users thinking it may be used by all servers

> [!NOTE]
> This will not compile fully - only playing w/ the contract a bit before going further

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/server/DefaultGrpcServerFactory.java
@@ -28,7 +28,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.grpc.internal.GrpcUtils;
-import org.springframework.lang.Nullable;
 
 import com.google.common.collect.Lists;
 import io.grpc.Grpc;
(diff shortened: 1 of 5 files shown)


dsyer:
I took this suggestion and extended it a bit in ac3ecb5. Also closes #216.

therepanic:
Hi, @onobc. You may have noticed that I deliberately did not include documentation in my commit, but it is needed. I can try to write it if you don't mind (I assume you noticed this and decided to do it yourself a little later). If so, can you advise which tab in the documentation to include it in? I can guess what it will look like, as there is documentation in previous changes with filters, and I have a rough idea of how to do it.

onobc:
Thanks for the reminder @therepanic . I will make sure that the client, server, and service filter docs all exist and are all up-to-date (reflect the latest changes @dsyer made).

onobc:
> Thanks for the reminder @therepanic . I will make sure that the client, server, and service filter docs all exist and are all up-to-date (reflect the latest changes @dsyer made).

The client/server interceptor docs were already updated by Dave to reflect reality.

I added a section for service filtering [in this PR](https://github.com/spring-projects/spring-grpc/pull/222) PTAL

PR #219, opened 2025-07-03
Minor refactor: Configures services for a gRPC server factory by discovering services and applying configuration.

NeatGuyCoding:
as stated in title


dsyer:
Could you squash it down to one commit, please (seems excessive to have 3 for such a small change)? Also please make a shorter title for the log, e.g. "Internal refactor to avoid repetition". Thanks.

Issue #207, opened 2025-06-28
Add ability to filter out services from being bound to server factories


onobc:
The main feature request from https://github.com/spring-projects/spring-grpc/issues/144 was solved in https://github.com/spring-projects/spring-grpc/commit/a087eac7764a22b9d89ee766283ddb0f17272e08. 
However,  one improvement would be the ability to filter out services from being bound to the InProcess server / factory.

Similarly to how we are [filtering out client side interceptors](https://github.com/spring-projects/spring-grpc/commit/161ad69de29b0f52b20d5373e48e333bfaa5d941) from channel factories, we should allow users to configure a filter(s) that does the same for the server-side factory.

therepanic:
Hi, @onobc! Since this feature is similar to what I implemented, can I try it too?

onobc:
Hi @therepanic ! I was hoping you were keeping your eye on the issue board (as you usually are 😺 - which is much appreciated). I reached out the team to get agreement on the need to do this filtering but I am 99% sure it is something we will all agree on. Feel free to submit a code proposal. 


> [!NOTE]
> While digging around to add some details to this issue, I also noticed that we will likely want to add filtering of the server-side global interceptors. I created #208 to cover this separately though.


### Details

I think we will want these service filtering components to live in the core module, like we did for the client filter components (namely `ClientInterceptorsConfigurer`). The `DefaultGrpcServerFactory#addService` method could do something like:

```java
  @Override
  public void addService(ServerServiceDefinition service) {
    if (this.serviceFilter == null || this.serviceFilter.filter(service, this)) {
      this.serviceList.add(service);
    }
  }
```

Wdyt?

therepanic:
Thanks.

Yes, I think that's okay, I will do that.

therepanic:
Hello @onobc again. I assume the team has approved everything? If so, I will send the PR tomorrow.

onobc:
Yes @therepanic - please go forward w/ this. Thank you.

therepanic:
To be honest, I don't quite understand how we can test this. I think we want to check that the filter has been applied and that we haven't added a `ServerServiceDefinition` instance to `DefaultGrpcServerFactory#serviceList` field. However, this field itself is private and there are no getters. On the other hand, I think it would be good to have a constructor for `serviceList`. I don't think it's a good idea to use reflection. Or is there another way?

onobc:
[...]

PR #222, opened 2025-07-04
Add docs for service filtering

onobc:
See #207

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/pages/server.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/pages/server.adoc
@@ -12,6 +12,37 @@ There are some `BindableServices` available off the shelf that you could include
 Very commonly, you will create your own `BindableService` by extending the generated service implementation from your Protobuf file.
 The easiest way to activate it is to simply add a Spring `@Service` annotation to the implementation class and have it picked up by the `@ComponentScan` in your Spring Boot application.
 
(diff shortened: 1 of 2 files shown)


Meijuh:
I can confirm service filtering is working. However, filtering out the "health service" and "reflection service" might be a more valuable example, if you agree not having them on in-process servers is good practice:

```java
return (serviceDefinition, __) ->
        !HealthGrpc.SERVICE_NAME.equals(serviceDefinition.getServiceDescriptor().getName())
            && !ServerReflectionGrpc.SERVICE_NAME.equals(
                serviceDefinition.getServiceDescriptor().getName())
```

Looking forward to a 0.9.0 release.

onobc:
> I can confirm service filtering is working. However, filtering out the "health service" and "reflection service" might be a more valuable example, if you agree not having them on in-process servers is good practice:
> 
> ```java
> return (serviceDefinition, __) ->
>         !HealthGrpc.SERVICE_NAME.equals(serviceDefinition.getServiceDescriptor().getName())
>             && !ServerReflectionGrpc.SERVICE_NAME.equals(
>                 serviceDefinition.getServiceDescriptor().getName())
> ```
> 
> Looking forward to a 0.9.0 release.

Good suggestion @Meijuh . PTAL at the result [here](https://docs.spring.io/spring-grpc/reference/server.html#service-filtering)

Meijuh:
LGTM!😀

PR #225, opened 2025-07-06
Drop old 'jakarta_omit' option

breun:
The `jakarta_option` was only supported by a single grpc-java release and quickly replaced by `@generated=omit` (https://github.com/grpc/grpc-java/pull/11086), which is already used by the sample configurations.

### DIFF
--- a/samples/grpc-client/build.gradle
+++ b/samples/grpc-client/build.gradle
@@ -51,7 +51,6 @@ protobuf {
     generateProtoTasks {
         all()*.plugins {
             grpc {
-                option 'jakarta_omit'
                 option '@generated=omit'
             }
         }
--- a/samples/grpc-client/pom.xml
+++ b/samples/grpc-client/pom.xml
@@ -113,7 +113,7 @@
 							<groupId>io.grpc</groupId>
 							<artifactId>protoc-gen-grpc-java</artifactId>
 							<version>${grpc.version}</version>
-							<options>jakarta_omit,@generated=omit</options>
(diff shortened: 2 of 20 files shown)


onobc:
Something else we need to be aware of if this change get approved, the generation of the option in start.spring.io ([details](https://github.com/spring-io/initializr/issues/1601))

dsyer:
Probably this change can be made independently though?

onobc:
> Probably this change can be made independently though?

Absolutely. Just noting the fact that it is also in start.spring.io. It might be safer to leave it as-is for a little bit to make sure this change does not cause issues. I remember it did initially cause a bit of grief (worked in Gradle but not Maven - or the other way around). 

That being said, did we test this change in both Gradle and Maven?

breun:
I created a pull request for start.spring.io: https://github.com/spring-io/start.spring.io/pull/1855

onobc:
> I created a pull request for start.spring.io: [spring-io/start.spring.io#1855](https://github.com/spring-io/start.spring.io/pull/1855)

Thanks @breun ! Do you mind verifying these changes w/ both Gradle and Maven if you did not already? As these options are consumed by the build plugins, I have previously seen inconsistent behavior between the two in this area.

Thanks

breun:
I have verified with Maven only, because that's what I use for my projects, but AFAIK these options just passed on to grpc-java, so it shouldn't matter whether you pass them via the Maven plugin or the Gradle plugin.

PR #226, opened 2025-07-06
Enable checkstyle on autoconfigure/docs modules

therepanic:
The motivation to do this came from @onobc's comment at https://github.com/spring-projects/spring-grpc/pull/223#issuecomment-3040989483. I think it would be worthwhile for us to include checkstyle at least in the autoconfigure module.

There were classes such as GrpcSecurityAutoConfiguration and OAuth2ResourceServerAutoConfiguration that contained non-static classes, which should not be done. In this case, I split them into new classes so that checkstyle would pass.

Also, the `GrpcClientAutoConfigurationTests` test is currently failing. I would like to ask if anyone has any ideas as to why this is happening.


spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/ConditionalOnGrpcClientEnabled.java
```diff
@@ -15,21 +15,23 @@
...
 /**
  * {@link Conditional @Conditional} that only matches when the {@code io.grpc:grpc-stub}
  * module is in the classpath and the {@code spring.grpc.client.enabled} property is not
  * explicitly set to {@code false}.
  *
- * @author Freeman
+ * @author Freeman Freeman
  * @author Chris Bono
  */
 @Retention(RetentionPolicy.RUNTIME)
```
therepanic, line 37: In the GitHub commit history, I couldn't find anyone named Freeman to find his last name, but checkstyle strictly requires that the last name be included.

spring-grpc-docs/pom.xml
```diff
@@ -19,7 +19,6 @@
 		<configprops.path>${project.basedir}/src/main/antora/modules/ROOT/partials/_configprops.adoc</configprops.path>
 		<configprops.inclusionPattern>spring.grpc.*</configprops.inclusionPattern>
 		<jruby.version>9.4.6.0</jruby.version>
-		<disable.checkstyle.checks>true</disable.checkstyle.checks>
```
dsyer, line 22: It seems to be this that causes the build failure. I don't see why this should be a blocker.

onobc: I think we can get this in after the release.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcChannelFactoryConfigurations.java
```diff
@@ -46,8 +46,7 @@
 class GrpcChannelFactoryConfigurations {
 
 	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnClass(value = { io.grpc.netty.shaded.io.netty.channel.Channel.class,
-			io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder.class })
+	@ConditionalOnClass({ Channel.class, NettyChannelBuilder.class })
```
onobc, line 49: The fully qualified package name is required as it is the path to the class the includes the "shaded" path (i.e. `NettyChannelBuilder.class != io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder.class`

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/security/GrpcSecurityAutoConfiguration.java
```diff
@@ -50,50 +31,3 @@
 public class GrpcSecurityAutoConfiguration {
```
onobc, line 31: What happened here? 😸

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcChannelFactoryConfigurations.java
therepanic: Yes, you are absolutely right, thank you!

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/security/GrpcSecurityAutoConfiguration.java
therepanic: As I described above, Checkstyle does not allow multiple classes in a single file, so I had to split them into separate files.

therepanic:
Thank you for your review. As you can see, most of the changes were made to the order of imports, the order of methods, and so on. Also, the order in javadoc. All of this accumulated over time, which is why so many changes had to be made.

I force-pushed because I accidentally touched pom.xml in the test module, which is not related to these changes. I also added a commit that fixes your comment. However, that test still fails.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/client/GrpcChannelFactoryConfigurations.java
```diff
@@ -72,7 +71,8 @@ ShadedNettyGrpcChannelFactory shadedNettyGrpcChannelFactory(GrpcClientProperties
 	}
 
 	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnClass(value = { Channel.class, NettyChannelBuilder.class })
+	@ConditionalOnClass({ io.grpc.netty.shaded.io.netty.channel.Channel.class,
```
onobc, line 74: This config is for the non-shaded NettyChannelFactory so this should stay non-fully-qualified and just add the import back in.

spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/security/GrpcSecurityAutoConfiguration.java
onobc: >There were classes such as GrpcSecurityAutoConfiguration and OAuth2ResourceServerAutoConfiguration that contained non-static classes, which should not be done. In this case, I split them into new classes so that checkstyle would pass.

I did not fully read the PR description (which is one of pet peeves when someone does not read mine 😸). My bad.

I understand why you split them. However, I prefer that we disable checkstyle for `GrpcSecurityAutoConfiguration` at the class level ([described below](https://github.com/spring-projects/spring-grpc/pull/226#pullrequestreview-2991665738)) rather than restructure the classes.

therepanic: I suppose restructuring is possible, but in another commit, since the current PR is aimed at fixing the module under checkstyle, and the best way to mark it is as you described below?

therepanic:
I had to do a rebase because after all the fixes, the code still failing due to https://github.com/spring-projects/spring-grpc/commit/f90cecccd5e3ead0d81c14435e8d2dd94f279e62.

In the last commit, I made all the corrections. I also disabled (at least temporarily) checkstyle on the doc module. I looked at what corrections needed to be made for checkstyle to approve the doc module and changed my mind. In order for checkstyle to pass in the doc module, we need to write an empty constructor in `ConfigurationPropertiesAsciidocGenerator` and replace all `System.out`.

What do you think about this? It seems to me that this is the very class for which we do not need to have an empty constructor. In any case, if we do checkstyle, we will also have to replace `System.out` with logging in another way.

Issue #231, opened 2025-07-16
OAuth2ResourceServerAutoConfiguration conflicts with ReactiveOAuth2ResourceServerAutoConfiguration


emil-e:
In my case, it's specifically the `JwtDecoder` named `jwtDecoderByIssuerUri` that conflicts with the `ReactiveJwtDecoder` with the same name from `ReactiveOAuth2ResourceServerAutoConfiguration`.

For now, I have this workaround (Kotlin):
```Kotlin
@Bean
fun jwtDecoder(reactiveJwtDecoder: ReactiveJwtDecoder) = JwtDecoder { reactiveJwtDecoder.decode(it).block() }
```

However, this still leaves with few options if I want to also disable the `AuthenticationProcessInterceptor` in order to provide a reactive one. Since the `OAuth2ResourceServerAutoConfiguration` in `spring-grpc` is an autoconfig root but not public I cannot easily disable it through `@ImportAutoConfiguration`. I could try to disable it using the `@ConditionalOnBean(JwtDecoder.class)` condition by not providing a `JwtDecoder`, but `spring-grpc` provides one anyway due to the presence of issuer URI config in my `application.yaml`. I don't want to remove the YAML config since that's where you're supposed to configure the `ReactiveOAuth2ResourceServerAutoConfiguration` too.

How is it supposed to work?

therepanic:
Hi, @emil-e. So, essentially, one solution to the problem would be to disable `OAuth2ResourceServerAutoConfiguration`, i.e. if it were a separate class with a file? In other words, auto-configuration would become public.

emil-e:
It has a separate file but if it was also public it could be referenced in `@ImportAutoConfiguration`, yes. The class in Spring Boot after which it was modeled _is_ public so I'm wondering if it's intentional that this is public - perhaps because it's copy/pasted from Spring Boot Security starter as a temporary (?) hack around the fact that the original configuration is not very reusable.

Making it public would be one option. Ideally, it should just disable itself more readily when there are conflicts. In the original auto config, the reactive vs. non-reactive configuration is exclusive to each other so this doesn't happen.

I'm unsure what the right solution is, honestly.

therepanic:
Yes, that's what I meant. In https://github.com/spring-projects/spring-grpc/pull/226#discussion_r2190760359, I wanted to separate the classes into different files instead of including the checkstyle, but in that particular PR, it wasn't a good idea because of the context of the PR itself. 

[...]

PR #233, opened 2025-07-17
Split multiple auto-config classes into separate files

therepanic:
I wanted to do this earlier https://github.com/spring-projects/spring-grpc/pull/226#discussion_r2190760359 but it didn't fit the context of the PR. However, I think we should still make this change. I don't see any reason why we should keep it as it is, since keeping classes in one file is bad practice.

Another trigger was https://github.com/spring-projects/spring-grpc/issues/231. I believe this fix can partially solve the problem; we can refer to the auto configurations and disable them. I can't say for sure how well it solves the problem, so I'm not attaching this issue to this PR.

I also removed the checkstyle tags from the auto configurations (//CHECKSTYLE:OFF/ON) as they are no longer needed. Without them, checkstyle complained that separate classes cannot be stored in one file. Accordingly, checkstyle will now be applied to each class in the files.

### DIFF
--- a/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/security/ExceptionHandlerAutoConfiguration.java
+++ b/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/security/ExceptionHandlerAutoConfiguration.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


therepanic:
I just saw https://github.com/spring-projects/spring-grpc/commit/866a7ecc6431fb3996a8ba8a2196c2a9adf1ea60. However, I don't think that's an obstacle.

onobc:
👋🏻  @therepanic 
Thanks for the proposal. I have assigned myself to review and will get to it asap.

dsyer:
It's -1 from me really for this change. The whole point of those classes was that they were package private and most likely to be retired/evolved when we integrate with Spring Boot, so we don't want people to be able to get hold of them. You might as well keep them all in the same place at that point IMO - it's neater.

onobc:
Thanks for taking the time to make the changes @therepanic . However, I agree w/ @dsyer as the

1. We are about to move all auto-configuration into Spring Boot and things will shift during that move.
2. Coarse grained auto-configuration w/ nested configurations is a common pattern in Spring Boot

I'm going to close this one. Again, thank you for all your contributions.

PR #239, opened 2025-07-28
Fix README: Add missing @GrpcService annotation to service examples

Shawyeok:
- Fixes issue where applications would exit immediately without `@GrpcService` annotation
- Update Spring Initializr dependency name from 'gRPC' to 'Spring gRPC'

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This section offers jumping off points for how to get started using Spring gRPC.
 
 Want to get started? Let’s speedrun a working service.
 
-Go to the [Spring Initializr](https://start.spring.io) and select the `gRPC` dependency.
+Go to the [Spring Initializr](https://start.spring.io) and select the `Spring gRPC` dependency.
 
 Generate the project and unzip the downloaded result.
 
@@ -74,7 +74,7 @@ Eclipse or VSCode will add them automatically for you.
 Now you can implement a service based on the generated stubs:
 
 ```java
-@Service
+@GrpcService
 class GrpcServerService extends SimpleGrpc.SimpleImplBase {
 
     private static Log log = LogFactory.getLog(GrpcServerService.class);
@@ -236,7 +236,7 @@ You need a Protobuf file that defines your service and messages, and you will ne
 Create a `@Bean` of type `BindableService`. For example:
 
 ```java
-@Service
+@GrpcService
 public class GrpcServerService extends SimpleGrpc.SimpleImplBase {
 ...
 }


onobc (commented):
Hi @Shawyeok ,

Thank you for the contribution. 

The top-level README is actually auto-generated from the sources in the `spring-grpc-docs` module as described [here](https://github.com/spring-projects/spring-grpc/tree/main/spring-grpc-docs#readme). I will be the first to raise my hand and say that this is confusing at first (I too have edited the file directly; multiple times 😸).

Do you mind moving these updates to [getting-started.adoc](https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-docs/src/main/antora/modules/ROOT/pages/getting-started.adoc#getting-started)?

You can then run `./build-docs.sh` locally to verify the generated top-level README.md.

Also, another point that is not clearly spelled out currently in the docs is that `@GrpcService` is optional and only required when you want to specify service interceptors. It is a composed annotation of `@Service`. The only true requirement is that a `BindableService` is found in the application context ([details here](https://github.com/spring-projects/spring-grpc/blob/88c190d6e99a28b5882707ab3d749c924de063b2/spring-grpc-core/src/main/java/org/springframework/grpc/server/service/DefaultGrpcServiceDiscoverer.java#L46)). Maybe it would be more helpful to clarify this in the docs than switch it in the getting started example. If we switched it in this sample then there are also many samples in `sammples` module that we would want to also switch.

Wdyt?

Thanks

Shawyeok:
@onobc 
Ah, I can’t reproduce the issue today. I misunderstood how `getOrderedBeansWithAnnotation` works — I thought it would return beans only if they were annotated with @GrpcService. Maybe copy-paste wasn’t my friend that day. 😅

> Also, another point that is not clearly spelled out currently in the docs is that @GrpcService is optional and only required when you want to specify service interceptors. It is a composed annotation of @Service. The only true requirement is that a BindableService is found in the application context ([details here](https://github.com/spring-projects/spring-grpc/blob/88c190d6e99a28b5882707ab3d749c924de063b2/spring-grpc-core/src/main/java/org/springframework/grpc/server/service/DefaultGrpcServiceDiscoverer.java#L46)). Maybe it would be more helpful to clarify this in the docs than switch it in the getting started example. If we switched it in this sample then there are also many samples in sammples module that we would want to also switch.

Thanks for clarification! Looks like the real mistake was mine.

Change dependency name from `gRPC` to `Spring gRPC` is trivial, therefore I'm going to close this.

onobc:
Thank you @Shawyeok .

PR #242, opened 2025-08-04
hotfix: fix global exception handler error log

NeatGuyCoding:


dsyer:
I'm pretty sure the code we have was intentional. Can you explain the change?

Issue #250, opened 2025-08-18
Make the default gRPC client stub factory configurable via application properties


ch200203:
**Summary**
When auto-registering gRPC client stubs, Spring gRPC currently defaults to a blocking stub factory. It would be helpful to make the global default stub factory configurable via `application.properties`/`application.yml`, aligning with Spring Boot’s convention-over-configuration approach. This lets users switch the default to a `coroutine`/`reactive`/`future` ... stub without additional bean wiring. The project already provides automatic client configuration and client properties for channels; this proposal adds one more property to control the default stub type.

**Motivation**
Projects often prefer a non-blocking model by default (e.g., Kotlin Coroutine stubs or Reactor stubs). Today, changing the default typically requires custom beans or explicit per-client configuration. A property-based global default would:

- Reduce boilerplate and improve DX
- Make behavior more discoverable
- Stay consistent with existing `spring.grpc.client.*` properties.

**Current Behavior**
Client auto-registration scans packages and registers client stubs with a default (blocking) stub factory, hard-coded in ClientScanConfiguration. There is no documented property to change this global default; users must override with custom beans or per-stub configuration.

**Code Reference (current)**
In `ClientScanConfiguration.java` , the default factory is hardcoded as BlockingStubFactory (note the inline TODO already hints at making this configurable):

```java
@Override
protected GrpcClientRegistrationSpec[] collect(AnnotationMetadata meta) {
    Binder binder = Binder.get(this.environment);
    boolean hasDefaultChannel = binder.bind("spring.grpc.client.default-channel", ChannelConfig.class)
        .isBound();
    if (hasDefaultChannel) {
        List<String> packages = new ArrayList<>();
        if (AutoConfigurationPackages.has(this.beanFactory)) {
            packages.addAll(AutoConfigurationPackages.get(this.beanFactory));
        }
        // TODO: change global default factory type in properties maybe?
        return new GrpcClientRegistrationSpec[] { GrpcClientRegistrationSpec.of("default")
            .factory(BlockingStubFactory.class)
            .packages(packages.toArray(new String[0])) };
    }
    return new GrpcClientRegistrationSpec[0];
}
```

**Proposed Change**
Introduce a new property to select the global default client stub factory:

```yml
# Suggested options
spring.grpc.client.default-stub-factory=blocking | async | future | coroutine | reactor #(sample)
# Optional escape hatch for custom factories (overrides the enum-like setting)
[...]

PR #252, opened 2025-08-21
Default stub factory configurable via properties (#250)

ch200203:
This PR makes the *global default gRPC client stub factory* configurable via Spring Boot properties. Instead of being hardcoded to the blocking stub factory during auto-registration, users can now choose any `StubFactory` by setting a single property.

**Property**

```properties
# FQCN of a StubFactory implementation
spring.grpc.client.default-stub-factory=org.springframework.grpc.client.ReactorStubFactory
```

* Type: `Class<? extends StubFactory<?>>` (per @dsyer’s suggestion — no string/enum alias map).
* Default: `org.springframework.grpc.client.BlockingStubFactory` (preserves current behavior).


dsyer:
Please run `mvn install` to check your changes.

dsyer:
You forgot to sign the last commit? Might be best to just squash them all down into a single commit and sign that.

ch200203:
Thanks for the heads-up! I missed signing the last commit—sorry about that. I’ll squash this PR down to a single commit, sign it, and push an update. 
@dsyer Thanks for your help! 🙇‍♂️

Issue #254, opened 2025-08-25
Expose ServerCall in GrpcAuthenticationExtractor#extract (enable method-aware auth)


jmezna:
I propose to change the signature of `GrpcAuthenticationExtractor#extract` from `GrpcAuthenticationExtractor#extract(io.grpc.Metadata, io.grpc.Attributes)` to `GrpcAuthenticationExtractor#extract(io.grpc.Metadata, io.grpc.ServerCall)`. 

This way the `Authentication` can be constructed using call details (e.g., method descriptor), not just headers/attributes.

dsyer:
The problem with that is that `ServerCall` gives you access to way more than just the `MethodDescriptor`, and it doesn't seem appropriate that an authentication extractor would need anything else. We could add a `MethodDescriptor` method parameter maybe. Can you describe a use case where you need it?

jmezna:
I am working with custom JWT tokens that are issued for a specific path (in servlet case) or a specific grpc method. 

When I construct the Authentication object, I want to capture the actual path/method that the request was sent to, so that I can later validate whether the token is valid for that resource. 

In the servlet world I can do this in `AbstractAuthenticationProcessingFilter#attemptAuthentication`, because I have access to the `HttpServletRequest` and can read the request path directly.

I’d like to have the same capability in the gRPC case — i.e., access to the `MethodDescriptor` when extracting authentication, so I can enforce that the JWT’s “audience” really matches the gRPC method being invoked.

dsyer:
That's helpful, thank you. An alternative (without API changes) would be to add an interceptor that packed the `MethodDescriptor` (or just the full method name) into one of the `Attributes`. That seems to fit the kind of patterns we see in Spring Web.

dsyer:
Another idea: we already have a `CallContext` which is a public record containing headers, attributes and method descriptor. We could use that in the authentication extractor too (also an API change).

therepanic:
Perhaps it would be better not to make API changes and come up with an implementation that doesn't require them? For example, your idea with the interceptor? In any case, I would like to work on this if you don't mind

dsyer:
It's OK to put an API change in before 1.0. Let's see if anyone else has an opinion - you are more than welcome to implement something in any case.

dsyer:
On reflection, and in the absence of any other input, I think I prefer to add a method parameter, so

```
[...]

PR #259, opened 2025-08-31
Add `MethodDescriptor` parameter to `GrpcAuthenticationExtractor#extract`

therepanic:
Closes: gh-254


therepanic:
From the https://github.com/spring-projects/spring-grpc/issues/254, we came to this solution.

Issue #269, opened 2025-09-23
Fix Observation property name


saurabhbhalla90:
Looks like the property to enable / disable observations in spring-grpc is called `spring.grpc.client.observations.enabled`

Reference: https://github.com/spring-projects/spring-grpc/blob/014123476f595c01eac0bf9946fac3cd9973f70f/spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc#L48

However the spring repo seems to be using `spring.grpc.client.observation.enabled` instead (observations vs observation)

Reference:
https://github.com/spring-projects/spring-grpc/blob/5b1ab080ed2d72682ec20bdec5f9747c478cc66e/spring-grpc-spring-boot-autoconfigure/src/main/java/org/springframework/grpc/autoconfigure/server/GrpcServerObservationAutoConfiguration.java#L36

onobc:
Hi @saurabhbhalla90 ,

Good catch. Luckily this is just a docs issue as the code all expects the singular `observation`.

A couple of things:

1. In the `main` branch the config props are moving into Spring Boot and we will be sure this gets fixed in my in progress PR to Boot.
2. This is an issue in the `0.x` branch and we should fix it there. Thanks @therepanic for the PR - I have added some details to it PTAL.

PR #270, opened 2025-09-23
Fix Observation property name in `_configprops.adoc`

therepanic:
In fact, this inaccuracy has been described since the time when the autoconfigure module was in spring-grpc. I think we should consider using the `observation` option, although both options are fine, but this option is older and more familiar to use (spring grpc 1.0.0 has not been released yet, but nevertheless). In addition, I think that the original intention was to use the `observation` option. Furthermore, there are already options in the Spring ecosystem that simply use `observation`, such as `spring.kafka.listener.observation-enabled`.

Closes: gh-269

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/partials/_configprops.adoc
@@ -23,7 +23,7 @@
 |spring.grpc.client.enabled | `+++true+++` | Whether to enable client autoconfiguration.
 |spring.grpc.client.inprocess.enabled | `+++true+++` | Whether to configure the in-process channel factory.
 |spring.grpc.client.inprocess.exclusive | `+++true+++` | Whether the inprocess channel factory should be the only channel factory available. When the value is true, no other channel factory will be configured.
-|spring.grpc.client.observations.enabled | `+++true+++` | Whether to enable Observations on the client.
+|spring.grpc.client.observation.enabled | `+++true+++` | Whether to enable Observations on the client.
 |spring.grpc.server.address |  | The address to bind to. could be a host:port combination or a pseudo URL like static://host:port. Can not be set if host or port are set independently.
 |spring.grpc.server.enabled | `+++true+++` | Whether to enable server autoconfiguration.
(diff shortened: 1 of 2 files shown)


onobc:
Agreed @therepanic, the singular `observation` is the intended name and is in fact the one used to guard the Observation auto-configuration. If we made it plural it would in fact be a breaking change as it is only a docs issue at this point.
m
> [!NOTE]
> The `_configprops.adoc` is actually generated via the config props processor in the docs module. The content for that file is aggregated from the config props POJOs and the [additional-spring-configuration-metadata.json](https://github.com/spring-projects/spring-grpc/blob/8216844d9ac34b29ac6efbeee63f01f998cf8afe/spring-grpc-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L39) file (here is where you define config props that are not defined in a config props POJO - e.g. these enabled properties). 

Here is what I think needs to happen:
- [x] update the generated file so we can re-run the docs Github workflow - this will fix the currently available docs for the `0.x` branch.
- [x] update the `additional-spring-configuration-metadata.json` file so that next time the docs are run that the generated file stays fixed?
- [x] re-target the PR to the `0.x` branch
- [x] fix this in Spring Boot for `main` branch

If you don't have time to deal w/ the `0.x` branch I am more than happy to cherry-pick your changes against main and put them in the `0.x` branch.

Sound good?

therepanic:
Hi, @onobc. Thank you for your review.

I did as you said and redirected the changes to `0.x`. Also, I assume you have already updated `additional-spring-configuration-metadata.json` here https://github.com/spring-projects/spring-boot/pull/47288/commits/6b2b95e17c30e17abb401d9f20fff9d288713bfc

onobc:
Hi @therepanic , do you mind doing the 2nd bullet item as well (updating the source file so going forward the next time we run the generated file will retain the proper property name)?

therepanic:
Sorry, I forgot that we are on the `0.x` branch and we need to update `additional-spring-configuration-metadata.json` in the autoconfigure module that is still in this branch. I rebase and push.

onobc (approved):
Thanks for the fix @therepanic ! LGTM

PR #289, opened 2025-11-10
Add boot docs back in

onobc:


dsyer:
We probably need to cherry pick 80acca86 (or copy the changes from there) as well

Issue #290, opened 2025-11-11
Update to Java 25 to match the rest of the portfolio


onobc:
- [ ] Keep the target and source at 17 but make sure we build w/ 25.
- [ ] Update the `check-samples.yml` GH workflow to include Java 17 and Java 25 for compatibility tests

onobc:
Closed via https://github.com/spring-projects/spring-grpc/commit/e8ad5b8a14fa6b134446b370af97d9465bcb75e4 and several other commits (listed above).

PR #296, opened 2025-11-16
Support Java 25

onobc:
This commit adds support for Java 25. Note that the java.version remains at 17 in order to force the target and source compatibility to be 17. However, if the running JVM is 25 it will be used to compile and run tests.

The Github workflows are also updated to run with Java 25.

**TODO**
- [ ] Make samples run w/ both 17 and 25

See #290


dsyer:
LGTM

PR #302, opened 2025-11-22
Use JSpecify for nullability constraints

### LINKED ISSUE #192: Migrate to JSpecify annotations for nullability constraints
See https://github.com/spring-projects/spring-framework/issues/28797

onobc:
Resolves #192

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/package-info.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/package-info.java
@@ -14,4 +14,10 @@
  * limitations under the License.
  */
 
+/**
+ * Core module for Spring gRPC.
+ */
+@NullMarked
 package org.springframework.grpc;
+
+import org.jspecify.annotations.NullMarked;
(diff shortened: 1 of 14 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/package-info.java
```diff
@@ -14,4 +14,10 @@
  * limitations under the License.
  */
 
+/**
+ * Core module for Spring gRPC.
+ */
+@NullMarked
 package org.springframework.grpc;
```
lobaorn-bitso: Hey @onobc I believe that actually every package needs to be annotated, right? Not only the top-level one, due to the limitations of how NullAway+JSpecify do things, its not inherited from the root to all sub packages.

For example, searching inside Spring Framework we have 356 mathces for files `package-info.java` with `NullMarked` inside:

`repo:spring-projects/spring-framework "@NullMarked" path:package-info.java`

https://github.com/search?q=repo%3Aspring-projects%2Fspring-framework+%22%40NullMarked%22+path%3Apackage-info.java&type=code

So to truly benefit from JSpecify I believe Spring gRPC still has some package-info to add, right?

therepanic: Hi, @onobc. If you don't mind or are unavailable, I can do PR for this myself.

PR #306, opened 2025-12-01
Avoid deprecated EpollEventLoopGroup usage

therepanic:
Replaces `EpollEventLoopGroup` with `MultiThreadIoEventLoopGroup` configured via `EpollIoHandler` factory.

This aligns with the Netty 5 IO handler architecture and avoids usage of deprecated epoll transport APIs.

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/NettyGrpcChannelFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/NettyGrpcChannelFactory.java
@@ -20,14 +20,16 @@ import java.util.List;
 
 import io.grpc.ChannelCredentials;
 import io.grpc.netty.NettyChannelBuilder;
+import io.netty.channel.MultiThreadIoEventLoopGroup;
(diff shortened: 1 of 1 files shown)


therepanic:
Btw (unrelated to the current change), for some reason I couldn't build the grpc-server-sample module (even without the current change), but when building with Actions, everything is fine. Here's part of the log: https://pastebin.com/7r1JQEMC

onobc:
> Btw (unrelated to the current change), for some reason I couldn't build the grpc-server-sample module (even without the current change), but when building with Actions, everything is fine. Here's part of the log: https://pastebin.com/7r1JQEMC

I will investigate separately - thanks for the heads up!

Hi @therepanic , so this works for me locally and I am not sure where in the world it is getting `ClientInterceptorsConfiguration` but that will fail because that config does not exist and I am not sure it ever did? Do you see it anywhere if you global search your Spring gRPC repo? Try cleaning out your .m2 local repo for org/springframework/grpc and see if that fixes it?

PR #314, opened 2025-12-03
Update client port configuration instructions

cppwfs:
Clarify that the `launched.grpc.port` environment variable must be set to specify the server port, rather than assuming port 9090. Add concrete example of setting the environment variable using export command.

### DIFF
--- a/samples/grpc-client/README.md
+++ b/samples/grpc-client/README.md
@@ -18,7 +18,10 @@ $ ./mvnw spring-boot:run
 message: "Hello ==> Alien"
 ```
 
-The server has to be running on port 9090. You can use the `grpc-server` sample for that.
+You will need to set the property `launched.grpc.port` to the port the server is running on.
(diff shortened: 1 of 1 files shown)


onobc:
Thanks @cppwfs for this great catch. 

So these samples are also serve as our integration tests.
We use Spring Boot testjars to launch the server used by the apps that are the target of the tests.
Those tests set the launched.grpc.port via testjars `@DynamicProperty(name = "launched.grpc.port", value = "port")` in the tests (see `org.springframework.grpc.sample.GrpcClientApplicationTests.ExtraConfiguration#grpcServer`).
When we made this change to `launched.grpc.port`, we forgot to think about just running it by hand (not via tests). 

I modified your change to use a default port in the sPEL thus allowing the user to set the port only when its not 9090. PTAL. 

Closing this in favor of https://github.com/spring-projects/spring-grpc/commit/b89ecd2e1a7df33eee07a97764d1211d3a46839b

Issue #234, opened 2025-07-18
Migrate autoconfiguration to Spring Boot


dsyer:
Should be mostly straightforward. Just a package change.

Things to watch out for:

* OAuth2 - autconfig classes were copied from Spring Boot, so there will be clashes and condition corner cases. It would be best if Spring Boot sorted its own approach out first - e.g. to make it easier to configure OAuth2 clients in a non-webapp. The biggest challenge for us will be gRPC embedded in a servlet or webflux app, and wanting to use OAuth2 as well (there should be some stuff that can be shared in the servlet case, but maybe not in webflux).
* Where to put the samples? There are probably too many for them all to be used as smoke tests in Spring Boot, but they are valuable integration tests. We could leave them here I guess (cyclic dependency), or move them to a new samples repo (like Spring Security).

ch200203:
I’d like to contribute to this issue and help get things started.

Since the initial task seems to be a relatively simple package change, I’d be happy to submit a small PR for that first.

Would it be okay if I start with a small PR?

HyunsangHan:
Hi 👋 
Can I take a look at this issue? :)

HyunsangHan:
Oh, sorry. I should have checked there was already an assignee.

onobc:
Hi @HyunSangHan ,

Thank you for reaching out. I am already about to start on this one and things can grow a bit out of hand when moving a project into Spring Boot (e.g. [spring-pulsar](https://github.com/spring-projects/spring-boot/pull/34763)). I know the spring-grpc issue list is down to a minimum (which is good and bad at the same time as it makes it difficult to find things to work on). We will discuss this at our next team meeting and see if we can find other candidate issues that may interest you. 

Thanks,
Chris

Sineaggi:
How will auto-configuration work with spring boot test?

As-is, if you want to test with `@SpringBootTest`, you'd set webEnvironment to your choice of mock, none or random_port. How would that interact with grpc? Will there be an equivalent grpcEnvironment, or will webEnvironment pull double-duty?

We have currently been using a combination of `properties = ["spring.grpc.server.port=0"],` and `@LocalGrpcPort`.

dsyer:
> if you want to test with `@SpringBootTest`

It shouldn’t be any different after the migration than it is now. If there’s something awkward about it please ask in a separate thread/issue. There are many examples in the project already.

onobc:
[...]

PR #327, opened 2025-12-16
Fix: Missing secure field in ChannelConfig.copy() method

VahidRamezaniDB:
The secure field is not being copied in the ChannelConfig.copy() method, which can lead to incorrect channel configuration when a channel config is copied after modifying the secure flag from its default value. For example while setting the `spring.grpc.client.default-channel.secure` to `false` and then call the `getChannel()` with an unknown name the secure field would still be set to `true` in the returned instance.

## Changes

* Added the field assignment to the copy method
* Modified the unit test to verify it

## Note
I also thought about how to enforce future extensions of the ChannelConfig to modify the copy method but I could think of nothing but using the `java.land.reflect` api which I currently think is not a preferred option in this project. But I will accept any other recomendations.

### DIFF
--- a/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientProperties.java
+++ b/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientProperties.java
@@ -361,6 +361,7 @@ public class GrpcClientProperties implements EnvironmentAware, VirtualTargets {
 			copy.maxInboundMetadataSize = this.maxInboundMetadataSize;
 			copy.userAgent = this.userAgent;
 			copy.defaultDeadline = this.defaultDeadline;
+			copy.secure = this.secure;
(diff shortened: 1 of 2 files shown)


onobc:
👋🏻 @VahidRamezaniDB 

Great 🐞 catch! Thanks for another great contribution.

Your [original commit](https://github.com/spring-projects/spring-grpc/commit/8e7bf8023d03ce6e0dc3db07aebca0da8727299f) closed this PR. I also added a minor [polishing commit](https://github.com/spring-projects/spring-grpc/commit/3017d77229372aa32bfaf530d6d1657e51c6b36d).

> I also thought about how to enforce future extensions of the ChannelConfig to modify the copy method but I could think of nothing but using the java.land.reflect api which I currently think is not a preferred option in this project. But I will accept any other recomendations.

Yes, I know that having the specific fields becomes a brittle maintenance point (this PR is the case and point of that). In general, Spring projects tend to avoid reflection for these cases and instead make sure we have proper test coverage. Another thing to note that we are likely to push back on any changes that are in the auto-configuration modules as they will move into Spring Boot 4.1 (https://github.com/spring-projects/spring-grpc/issues/234).

Issue #329, opened 2025-12-17
OAuth2 resource server should not need to be a client


dsyer:
The oauth2 sample uses client credentials grant, and for that case you *do* need a client repository, but in general you do not (I guess, unless it's wrong in the Spring Boot modules as well). So probably there's a bug in the autoconfiguration conditions.

therepanic:
Hi, @dsyer.

Could it be that our `OAuth2ResourceServerAutoConfiguration` has the `InMemoryClientRegistrationRepository` class from the client module in its `@ConditionalOnClass`? Should we remove it?

I'm willing to submit a PR if that's ok

dsyer:
Yes, that's the problem. I'm not sure if simply removing it will fix the problem though - it must have been there for a reason (some resources servers need a client repository, as is the case with our oauth2 sample). Feel free to experiment.

therepanic:
I attached the PR and left a comment stating my thoughts on the matter. PTAL https://github.com/spring-projects/spring-grpc/pull/331.

In short, our grpc-oauth sample itself uses both the resource server and client modules directly, and uses both parts. If we remove this dependency, nothing will change. In addition, I believe that the auto-configuration of the resource server should not depend on the client part. This looks like some kind of error, maybe you have some thoughts on this.

PR #331, opened 2025-12-17
Fix resource server autoconfig to not require OAuth2 client

therepanic:
In this commit, we remove the `InMemoryClientRegistrationRepository` class from the `OAuth2ResourceServerAutoConfiguration` clause, after which we become independent of the OAuth2 client side.

Closes: gh-329

### DIFF
--- a/spring-grpc-server-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/server/autoconfigure/security/OAuth2ResourceServerAutoConfiguration.java
+++ b/spring-grpc-server-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/server/autoconfigure/security/OAuth2ResourceServerAutoConfiguration.java
@@ -52,7 +52,6 @@ import org.springframework.grpc.server.security.AuthenticationProcessInterceptor
 import org.springframework.grpc.server.security.GrpcSecurity;
 import org.springframework.security.config.Customizer;
 import org.springframework.security.config.ObjectPostProcessor;
-import org.springframework.security.oauth2.client.registration.InMemoryClientRegistrationRepository;
 import org.springframework.security.oauth2.core.DelegatingOAuth2TokenValidator;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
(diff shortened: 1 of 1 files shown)


spring-grpc-server-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/server/autoconfigure/security/OAuth2ResourceServerAutoConfiguration.java
```diff
@@ -88,8 +88,7 @@
 		after = { GrpcSecurityAutoConfiguration.class, GrpcServerFactoryAutoConfiguration.class })
 @EnableConfigurationProperties(OAuth2ResourceServerProperties.class)
 @ConditionalOnSpringGrpc
-@ConditionalOnClass({ InMemoryClientRegistrationRepository.class, BearerTokenAuthenticationToken.class,
-		ObjectPostProcessor.class })
+@ConditionalOnClass({ BearerTokenAuthenticationToken.class, ObjectPostProcessor.class })
```
therepanic: As stated here, this class was copied:
```java
// All copied from Spring Boot
// (https://github.com/spring-projects/spring-boot/issues/43978), except the
// 2 @Beans of type AuthenticationProcessInterceptor
```
However, to be honest, I didn't quite figure out where exactly it came from. Therefore, I don't quite understand why the `@ConditionalOnClass` class `InMemoryClientRegistrationRepository` was inserted. In fact, the server part should be completely isolated from the client part. In addition, the deletion was completely successful and all tests passed.

I also wrote a small local test (I didn't attach it because I thought it wouldn't be entirely relevant) that somehow checks whether we are isolated from the client part when starting the server resource. I did something like this:
```java
private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
    .withClassLoader(new FilteredClassLoader(“org.springframework.security.oauth2.client”))
    .withConfiguration(
        AutoConfigurations.of(GrpcSecurityAutoConfiguration.class, OAuth2ResourceServerAutoConfiguration.class))
    .withUserConfiguration(TestConfig.class);

```
And everything went well, in principle. Without this change, the test given in main fails.

Issue #334, opened 2025-12-22
Feature Request: Enable gRPC Health service without BindableService beans


nopan-rataullin:
**Feature request**: Allow gRPC Health service creation without existing BindableService

**Problem**
Services that don't yet have their own `@GrpcService` implementations don't get the gRPC Health service because of this [condition](https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-server-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/server/autoconfigure/health/GrpcServerHealthAutoConfiguration.java#L64):

  ```java
  @ConditionalOnBean(BindableService.class)
```

  Could you add an option to create the gRPC Health service without requiring an existing BindableService?

  

  Environment

  - Spring gRPC: 0.9.0
  - Spring Boot: 3.5.9

dsyer:
I'm not sure I understand your premise. What's the point of a service that doesn't have a registered `BindableService`?

nopan-rataullin:
@dsyer All our services are created from a gRPC template, and for consistency 
we want all of them to implement health checks via gRPC that can be 
integrated with Spring Actuator.

A quick dirty hack would be to implement a @GrpcService that does nothing
It's an edge solution for an edge case

dsyer:
I might have to see some actual code to understand what you are doing. I'm not sure what you mean by a "template". Are your services somehow not gRPC services, and yet you want to register them with the health service and actuators?

nopan-rataullin:
@dsyer Yes, exactly. Our primary method of communication between microservices is gRPC. Some services have @GrpcService implementations, while others only client-side. To avoid writing different configs for each service’s health check, it would be convenient to have a single, consistent gRPC health check endpoint that integrates with Spring Actuator for all our services

nopan-rataullin:
Let me also explain why integration with Spring Actuator is needed. For example, to extend the standard health checks with a check for the availability of a remote resource (not in our env) - if the resource is unavailable, the health check would fail

Otherwise, we’d end up using an webserver it in some cases and a gRPC check endpoint for services that have @GrpcService, which isn’t very consistent
Or use HTTP endpoint everywhere, which adds extra overhead for services that already have a GrpcServer

nopan-rataullin:
@dsyer Hope I explained that clearly
Let me know if you think my approach is wrong

dsyer:
[...]

PR #335, opened 2025-12-23
Refine default enablement of gRPC health service

therepanic:
Remove the `BindableService` bean condition from gRPC health auto-configuration. Make server-side health opt-in by default `spring.grpc.server.health.enabled=false`. This allows exposing the health service as the only gRPC service without starting a server unexpectedly unless enabled.

Closes: gh-334

### DIFF
--- a/spring-grpc-docs/src/main/antora/modules/ROOT/pages/health.adoc
+++ b/spring-grpc-docs/src/main/antora/modules/ROOT/pages/health.adoc
@@ -7,6 +7,8 @@ The health service is registered with the gRPC server and a `HealthStatusManager
 
 IMPORTANT: The health service resides in the `io.grpc:grpc-services` library which is marked as `optional` by Spring gRPC. You must add this dependency to your application in order for it to be autoconfigured.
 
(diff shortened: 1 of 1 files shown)


therepanic:
I will also add documentation where I will describe that server-side health in our case is disabled by default.

onobc:
👋🏻 @therepanic , thanks for the quick turnaround. However, we would like to have a team chat about this change before proceeding. We will get back w/ you shortly.

dsyer:
It's more complicated than my comment in #334 implied. We don't want the health checks to be disabled by default if there actually *is* a `BindableService`. So we need a composite condition with split logic for when there is or isn't a `BindableService` registered.

therepanic:
You mean it should work like this?

- If the application has at least one `BindableService`, then health should be enabled by default, as before.
- If the `BindableService` bean isn't registered in the application, then it should be disabled by default and only enabled if we specify it in the config?

dsyer:
Precisely

therepanic:
Okay, I'll be back soon with some polished changes.

therepanic:
I've slightly reworked my changes. They now incorporate the logic described above. I've also documented it.

Issue #328, opened 2025-12-17
Explain why GrpcSecurity is not available in a servlet app


dsyer:
Users can be confused when `GrpcSecurity` is not available because they have a `GrpcServlet` (they should use the normal web MVC security in that case). The best thing to do is probably provide a failure analyzer that recognizes the scenario where user tries to inject `GrpcSecurity` but it isn't available despite being on the classpath.

therepanic:
Hi, @dsyer. I suppose we could add something like custom `AbstractFailureAnalyzer` to the autoconfigure module and implement the corresponding logic in our scenario?
I can work on this. Could you assign me?

PR #340, opened 2026-01-02
Add FailureAnalyzer for missing GrpcSecurity in servlet mode

therepanic:
Provide a clear startup failure analysis when `GrpcSecurity` is requested while the gRPC server is running in servlet mode.

Closes: gh-328

### DIFF
--- a/spring-grpc-server-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/server/autoconfigure/GrpcSecurityInServletFailureAnalyzer.java
+++ b/spring-grpc-server-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/server/autoconfigure/GrpcSecurityInServletFailureAnalyzer.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2012-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


therepanic:
I am also open to any suggestions regarding messages in `getDescription` and `getAction`, perhaps you can describe them in more detail, idk.

spring-grpc-server-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/server/autoconfigure/GrpcSecurityInServletFailureAnalyzer.java
```diff
@@ -0,0 +1,81 @@
...
+		this.environment = environment;
+	}
+
+	@Override
+	protected @Nullable FailureAnalysis analyze(Throwable rootFailure, NoSuchBeanDefinitionException cause) {
+		if (!isMissingGrpcSecurity(cause) || !isActuallyServletMode()) {
+			return null;
+		}
+		return new FailureAnalysis(getDescription(), getAction(), cause);
+	}
```
therepanic: We want to call the current analyzer if and only if the thrown exception is related to `GrpcSecurity` and when the `spring.grpc.server.servlet.enabled` option is enabled and, accordingly, `GrpcServlet` is present in the path.

BTW, I'm not entirely sure whether `GrpcServlet` is always in the path or not, you can correct me if this is not the case, then we can simplify the check in `isActuallyServletMode` a little.

Issue #279, opened 2025-10-31
Function request: The client supports cluster load balancing


KouShenhai:
**Function request: The client supports cluster load balancing**

**description**

Generally, projects will be deployed in clusters in production environments, and we hope to support cluster load balancing

**problem**

Does it currently support client cluster load balancing?
If not, are there any plans to add this feature in the future?

Thank you for your consideration.

dsyer:
Whatever features are in grpc-java are available now as far as I know. If you want integration with Spring Cloud load balancers that's on the list of things we might do after 1.0.

KouShenhai:
@dsyer Thank you very much for your reply. Yes, we need to migrate Spring Cloud OpenFeign to gRpc.

dsyer:
Spring Cloud OpenFeign does have client-side load balancing, but so does grpc-java (just not using Spring Cloud). Please try the existing native features and let us know what the experience is like. You are the first to ask for Spring Cloud integration, so it will be interesting to see how valuable it might be.

KouShenhai:
> Spring Cloud OpenFeign does have client-side load balancing, but so does grpc-java (just not using Spring Cloud). Please try the existing native features and let us know what the experience is like. You are the first to ask for Spring Cloud integration, so it will be interesting to see how valuable it might be.

@dsyer Can I submit a Spring Cloud load balancer PR for this project? Do I need to create a Spring Cloud module before submitting? Could you give me some advice? Thanks.

dsyer:
You definitely can do that (the new module). Ultimately I’d expect those features to live in Spring Cloud, but we can host them here to start with.

PR #343, opened 2026-01-10
Add Spring Cloud load balancing to the client

KouShenhai:


dsyer:
As already mentioned in #279 we can’t take a Spring Cloud dependency here. It would be a good idea to submit this feature as a pull request or some pull requests in Spring Cloud.

KouShenhai:
@dsyer Yes, which module should I submit the code to in Spring Cloud? Do you have any suggestions?

spencergibb:
spring Cloud commons is the project where load balancer lives

KouShenhai:
@spencergibb thank

KouShenhai:
@spencergibb I don't understand, can't we create a separate project like Spring Cloud Openfeign? I feel like there's a lot of code that needs to be written in this, so it's actually better to make it a separate project

spencergibb:
No, we won't be creating new projects.

PR #344, opened 2026-01-15
Support named in-process channels

### LINKED ISSUE #342: `CompositeGrpcChannelFactory` identifies default inprocess target as supported by factories other than `InProcessGrpcChannelFactory`
As part of identifying the appropriate factory to use for a given target, `CompositeGrpcChannelFactory` finds the first factory for which supports(target) returns true.

```
@Override
	public ManagedChannel createChannel(final String target, ChannelBuilderOptions options) {
		return this.channelFactories.stream()
			.filter((cf) -> cf.supports(target))
			.findFirst()
			.orElseThrow(
					() -> new IllegalStateException("No grpc channel factory found that supports target : " + target))
			.createChannel(target, options);
	}
```

When the target is the default channel configured via application properties, `target` in the above method will evaluate to `default`. If the default channel address has been prefixed with `in-process`, the above code incorrectly identifies `target` as being supported by factories other than `InProcessGrpcChannelFactory`.

```
grpc:
    client:
      default-channel:
        address: in-process:test
```

As a result, the createChannel call ultimately fails with an IllegalArgumentException due to having an invalid DNS name.

```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [java.lang.Object]: Factory method 'getClient' threw exception with message: Invalid DNS name: in-process:test
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.lambda$instantiate$0(SimpleInstantiationStrategy.java:183)
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiateWithFactoryMethod(SimpleInstantiationStrategy.java:72)
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:152)
	at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:653)
	... 39 more
Caused by: java.lang.IllegalArgumentException: Invalid DNS name: in-process:test
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:217)
	at io.grpc.internal.DnsNameResolver.<init>(DnsNameResolver.java:168)
	at io.grpc.internal.DnsNameResolverProvider.newNameResolver(DnsNameResolverProvider.java:64)
	at io.grpc.internal.ManagedChannelImpl.getNameResolver(ManagedChannelImpl.java:682)
	at io.grpc.internal.ManagedChannelImpl.<init>(ManagedChannelImpl.java:605)
	at io.grpc.internal.ManagedChannelImplBuilder.build(ManagedChannelImplBuilder.java:731)
	at io.grpc.ForwardingChannelBuilder2.build(ForwardingChannelBuilder2.java:278)
	at org.springframework.grpc.client.DefaultGrpcChannelFactory.createChannel(DefaultGrpcChannelFactory.java:120)
```

onobc:
This commit allows in-process channels to be configured by name in application properties. Previously, the target (not the virtual target / address) was used to determine if the channel refers to an in-process one (via 'in-process:' prefix).

Fixes #342

### DIFF
--- a/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/ClientPropertiesChannelBuilderCustomizer.java
+++ b/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/ClientPropertiesChannelBuilderCustomizer.java
@@ -48,11 +48,11 @@ class ClientPropertiesChannelBuilderCustomizer<T extends ManagedChannelBuilder<T
 	}
 
 	@Override
-	public void customize(String authority, T builder) {
-		ChannelConfig channel = this.properties.getChannel(authority);
+	public void customize(String target, T builder) {
+		ChannelConfig channel = this.properties.getChannel(target);
 		PropertyMapper mapper = PropertyMapper.get();
 		mapper.from(channel.getUserAgent()).to(builder::userAgent);
-		if (targetAllowsLoadBalancer(authority)) {
+		if (targetAllowsLoadBalancer(target)) {
 			mapper.from(channel.getDefaultLoadBalancingPolicy()).to(builder::defaultLoadBalancingPolicy);
 		}
 		mapper.from(channel.getMaxInboundMessageSize()).asInt(DataSize::toBytes).to(builder::maxInboundMessageSize);
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
@@ -81,7 +81,7 @@ public class DefaultGrpcChannelFactory<T extends ManagedChannelBuilder<T>>
 
 	/**
 	 * {@inheritDoc}
-	 * @param target the target string as described in method javadocs
+	 * @param target the target string
 	 * @return true unless the target begins with 'in-process:'
 	 */
 	@Override
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelBuilderCustomizer.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/GrpcChannelBuilderCustomizer.java
@@ -18,11 +18,13 @@ package org.springframework.grpc.client;
 
 import java.util.function.Consumer;
 
+import io.grpc.ChannelCredentials;
+import io.grpc.Grpc;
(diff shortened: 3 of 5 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -112,10 +114,10 @@ public ManagedChannel createChannel(String target, ChannelBuilderOptions options
 		this.interceptorsConfigurer.configureInterceptors(builder, options.interceptors(),
 				options.mergeWithGlobalInterceptors(), this);
 		// Handle customizers
-		this.globalCustomizers.forEach((c) -> c.customize(target, builder));
+		this.globalCustomizers.forEach((c) -> c.customize(targetUri, builder));
```
onobc: The customizers do not expect "named" channels but rather the authority string or uri for the target.

spring-grpc-core/src/main/java/org/springframework/grpc/client/DefaultGrpcChannelFactory.java
```diff
@@ -80,15 +80,17 @@ public DefaultGrpcChannelFactory(List<GrpcChannelBuilderCustomizer<T>> globalCus
...
-	 * valid nameresolver-compliant URI or an authority string as described in
-	 * {@link Grpc#newChannelBuilder(String, ChannelCredentials)}.
+	 * {@inheritDoc}
 	 * @param target the target string as described in method javadocs
 	 * @return true unless the target begins with 'in-process:'
 	 */
 	@Override
 	public boolean supports(String target) {
-		return !target.startsWith("in-process:");
+		if (target.startsWith("in-process:")) {
```
onobc: We now check the target and the virtual target for the prefix.

spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/ClientPropertiesChannelBuilderCustomizer.java
```diff
@@ -75,6 +75,12 @@ public void customize(String authority, T builder) {
 		}
 	}
 
+	private boolean targetAllowsLoadBalancer(String authority) {
+		var targetUri = this.properties.getTarget(authority);
```
onobc: And yet again, we have to check both the target and targetUri as the prefixed uri could be either the "named" channel or the virtual target.

PR #347, opened 2026-01-16
Allow named channels to inherit default configuration

### LINKED ISSUE #345: Allow named channels to inherit configuration from default-channel
### Problem

Currently, when a channel is explicitly configured in `spring.grpc.client.channels.*`, it does not inherit any settings from `spring.grpc.client.default-channel.*`.

Looking at [`GrpcClientProperties.getChannel()`](https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientProperties.java#L100):

```Java
public ChannelConfig getChannel(String name) {
    if ("default".equals(name)) {
        return this.defaultChannel;
    }
    ChannelConfig channel = this.channels.get(name);
    if (channel != null) {
        return channel;  // Returns directly, no inheritance from defaultChannel
    }
    // defaultChannel.copy() is only used for channels NOT in the map
    channel = this.defaultChannel.copy();
    // ...
}
```
### Use Case

When configuring multiple gRPC clients, I want to set common properties once in `default-channel`/`global`  and have all named channels inherit them:
```YAML
spring.grpc.client:
  default-channel:
    max-inbound-message-size: 5MB
    max-inbound-metadata-size: 1MB
    enable-keep-alive: true
  channels:
    service-a:
      address: service-a:6565
      # Should inherit max-inbound-message-size, max-inbound-metadata-size, enable-keep-alive
    service-b:
      address: service-b:6565
      # Should inherit from default-channel too### Current Behavior
```
Each named channel must explicitly define all properties, leading to repetition:
```YAML
spring.grpc.client.channels:
  service-a:
    address: service-a:6565
    max-inbound-message-size: 5MB
    max-inbound-metadata-size: 1MB
  service-b:
    address: service-b:6565
    max-inbound-message-size: 5MB  # Duplicated
    max-inbound-metadata-size: 1MB  # Duplicated### Current Workaround
```
Use `GrpcChannelBuilderCustomizer` bean to apply settings globally:
```Java
@Bean
GrpcChannelBuilderCustomizer<?> globalCustomizer() {
    return (name, builder) -> builder
        .maxInboundMetadataSize(1024 * 1024)
        .maxInboundMessageSize(5 * 1024 * 1024);
}This works but requires code instead of configuration.
```

### Proposed Solution

Named channels should merge/inherit from `default-channel`:
```Java
public ChannelConfig getChannel(String name) {
    if ("default".equals(name)) {
        return this.defaultChannel;
    }
    ChannelConfig channel = this.channels.get(name);
    if (channel != null) {
        return this.defaultChannel.mergeWith(channel);  // Merge with defaults
    }
    channel = this.defaultChannel.copy();
    // ..
}
```
PR: https://github.com/spring-projects/spring-grpc/pull/347
This would be consistent with how other Spring Boot configurations handle inheritance/defaults.

o-shevchenko:
Named channels now inherit settings from default configuration. Explicitly configured values in named channels take precedence over defaults.

[resolves #345]

### DIFF
--- a/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientProperties.java
+++ b/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientProperties.java
@@ -20,11 +20,13 @@ import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Consumer;
 
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.boot.convert.DurationUnit;
 import org.springframework.context.EnvironmentAware;
 import org.springframework.core.env.Environment;
@@ -53,6 +55,11 @@ public class GrpcClientProperties implements EnvironmentAware, VirtualTargets {
 	private final Map<String, ChannelConfig> channels = new HashMap<>();
 
 	/**
+	 * Default configuration that named channels can inherit from.
+	 */
+	private final Channel channel = new Channel();
+
+	/**
 	 * The default channel configuration to use for new channels.
 	 */
 	private final ChannelConfig defaultChannel = new ChannelConfig();
@@ -71,6 +78,10 @@ public class GrpcClientProperties implements EnvironmentAware, VirtualTargets {
 
 	public Map<String, ChannelConfig> getChannels() {
 		return this.channels;
+	}
+
+	public Channel getChannel() {
+		return this.channel;
 	}
 
 	public ChannelConfig getDefaultChannel() {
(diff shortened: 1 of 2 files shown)


onobc (changes requested):
Thank you for another awesome contribution @o-shevchenko . I have left a few comments, other than that we are close to merge.

onobc:
Thanks for the great contribution @o-shevchenko , nice work! I did add a follow-on commit (https://github.com/spring-projects/spring-grpc/commit/d2266d7ad12e4cb95365be443477aa2652116535) which applies the defaults to the following channels when their `inherit-defaults` is `true`:

- default channel
- named channels
- unknonwn channels (copy of default)

Also, moved the defaults back to code w/ the "diff detection" being "did the user change the value from the coded default in the class". See my commit msg for reasoning. 

Again, since this is about to go into Spring Boot, and all auto-config / props will be under scrutiny I wanted to get this in asap. Otherwise I would have done more back/forth to get your opinion on my recent updates. 

Please review and we can address any of your concerns in subsequent movements. 

Thanks,
Chris

Closing via https://github.com/spring-projects/spring-grpc/commit/0bc8c88bc21c0c77e03bd1affb591bc62432ee56.

o-shevchenko:
Thanks @onobc !

PR #349, opened 2026-01-17
Add @GrpcClient annotation for field injection

### LINKED ISSUE #348: Add @GrpcClient annotation for simplified client injection
# Feature: Add @GrpcClient annotation for simplified client injection

## Summary

Add a `@GrpcClient` annotation to simplify gRPC client injection, similar to `@GrpcService` on the server side.
## Current approach:
```Java
private final ReactorServiceGrpc.ReactorServiceStub client;

    public MyGrpcClient(GrpcChannelFactory grpcChannelFactory) {
        this.client = ReactorServiceGrpc.newReactorStub(
                grpcChannelFactory.createChannel("client")
        );
    }
```

## Proposed approach
```Java
@GrpcClient("client")
private ReactorServiceGrpc.ReactorServiceStub client;
```

- Less boilerplate
- Per-client interceptors via annotation
- Mirrors `@GrpcService` pattern
- Backward compatible with existing approaches

PR: https://github.com/spring-projects/spring-grpc/pull/349

o-shevchenko:
Introduce @GrpcClient annotation to simplify gRPC client creation via field and setter injection. This provides an alternative to manual constructor injection with GrpcChannelFactory.

Signed-off-by: Oleksandr Shevchenko <shevchenko.olexandr96@gmail.com>

[resolves #348]

### DIFF
--- a/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java
+++ b/spring-grpc-client-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java
@@ -32,6 +32,7 @@ import org.springframework.grpc.client.ChannelCredentialsProvider;
 import org.springframework.grpc.client.ClientInterceptorsConfigurer;
 import org.springframework.grpc.client.CoroutineStubFactory;
 import org.springframework.grpc.client.GrpcChannelBuilderCustomizer;
+import org.springframework.grpc.client.GrpcClientBeanPostProcessor;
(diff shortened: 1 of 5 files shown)


dsyer:
Field injection in Spring is only ever done with `@Autowired`, so while the convenience factor is high here, we can't merge this in it's current form because it goes the other direction. It would have to work better with the existing `@ImportGrpcClients` approach, and use Spring's existing autowire support based on `@Qualifier`. Those are the patterns we have, and the implementation machinery supports them (as it does in parallel in all the other Spring projects, like HTTP clients in Spring Core and Spring Cloud).

o-shevchenko:
I'm closing this PR since the approach with `@Qualifier` sounds more useful and native for Spring

See https://github.com/spring-projects/spring-grpc/issues/348#issuecomment-3771886038

PR #371, opened 2026-02-05
Make SecurityGrpcExceptionHandler ordered

skjolber:
The default `GrpcExceptionHandlerInterceptor` sorts `GrpcExceptionHandler` by `Ordered`. Non-ordered are assigned `Ordered.LOWEST_PRECEDENCE`.

Use-case: Write handler for `RuntimeException` which runs after `SecurityGrpcExceptionHandler`.

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/server/security/SecurityGrpcExceptionHandler.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/server/security/SecurityGrpcExceptionHandler.java
@@ -20,6 +20,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.core.Ordered;
 import org.springframework.grpc.server.exception.GrpcExceptionHandler;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.core.AuthenticationException;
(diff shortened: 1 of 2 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/server/security/SecurityGrpcExceptionHandler.java
```diff
@@ -20,17 +20,24 @@
...
 
 import io.grpc.Status;
 import io.grpc.StatusException;
 
-public class SecurityGrpcExceptionHandler implements GrpcExceptionHandler {
+public class SecurityGrpcExceptionHandler implements GrpcExceptionHandler, Ordered {
 
 	private static final Log logger = LogFactory.getLog(SecurityGrpcExceptionHandler.class);
 
+	private final int order;
```
onobc: I think it may be more flexible if we instead move the ordering to the autoconfiguration (i.e. annotate the bean method w/ ordered annotation [example](https://github.com/spring-projects/spring-grpc/commit/54a8c3451fadb67536a778b600817648434442fe)). 

Wdyt?

onobc:
Thank you for the contribution! Nice work.

I was waiting on your reply to my PR review comments. However, for your to reply to my comments, I actually have to submit the review LOL.  

Screenshot 2026-02-12 at 10 53 03 AM

They were minor comments (the other one was around adding a test for the ordering). I went ahead and just implemented these changes in my [polishing commit](https://github.com/spring-projects/spring-grpc/commit/c0e6d56930c8cd93311d225ba09adebc841c133b).

skjolber:
Great, thank you @onobc! 🥇

PR #373, opened 2026-02-15
Support for Hot TLS Certificate Rotation

### LINKED ISSUE #49833: Support for gRPC server TLS certificate rotation
### Description
Is there currently support for hot TLS certificate rotation in Spring gRPC? Hot TLS certificate rotation refers to the ability to update/replace TLS certificates without restarting the gRPC server.

### Use Case
In production environments, TLS certificates need to be rotated periodically for security reasons or when they expire. Restarting servers to apply new certificates causes service disruption. Hot certificate rotation would allow for certificate updates while maintaining service availability.

### Current Understanding
Based on reviewing the codebase and documentation:
- Spring gRPC supports TLS configuration through Spring Boot's SSL bundle configuration
- Custom TrustManager implementations are possible (as documented in detail)
- However, there doesn't appear to be a mechanism to reload/rotate certificates at runtime

### Potential Implementation Approaches
If this feature doesn't exist yet, potential approaches might include:
1. A certificate watcher that monitors certificate files for changes
2. An API to programmatically update certificates at runtime
3. Integration with Spring Boot's SSL bundle refresh mechanisms (if available)

### Question
1. Is hot TLS certificate rotation currently supported?
2. If not, are there plans to add this feature in the future?
3. Are there any workarounds or recommended approaches for certificate rotation without service disruption?

Thank you for your consideration.

ukasus:
The following changes were made:
- Introduced `ReloadableX509KeyManager` and `ReloadableX509TrustManager` to wrap the dynamic gRPC SSL managers.
- Added 'GrpcServerSslManagerResolver' to resolve the reloadable managers from sslBundles and bind them to the `addBundleUpdateHandler` handler.
- Using set injection for setting managers in defaultGrpcServerFactory for all server factories.

Signed-off-by: Ujjawal Sharma <ujjawal98kaushik@gmail.com>

[resolves spring-projects/spring-boot#49833]


dsyer:
Thanks for the patch. It seems grpc-java doesn't support live SSL context reload the way that Tomcat and Jetty do, so it would be good if they were to change that, but I'm not holding my breath. Consequently I see why you have to introduce reloadable extensions of `KeyManager` and `TrustManager`. What I don't really understand is why you can't just inject those directly into the credentials in the `GrpcServerFactory` (there's no need in that case for all the extra `@Beans` in this patch, or the levels in between there and the server factory). I suspect, if you have one, your AI friend has been over enthusiastic, and might need to be prompted a bit to reduce the amount of clutter in this new code.

ukasus:
Thanks @dsyer for your valuable time and feedback, refactored the code as per your inputs.

onobc:
Let's wait until spring-projects/spring-grpc#234 is merged into Spring Boot to push this through as the auto-configuration will go away in 1.1 once it moves to Spring Boot (i.e. this will be a PR in Spring Boot).

spencergibb:
Watch https://github.com/spring-projects/spring-boot/issues/49043

philwebb:
Given the SSL code has now moved to Spring Boot I don't think we can easily migrate this PR, I'm going to close this one. Thanks anyway @ukasus! 

If we try again in Boot, I'm not sure that we want to adopt the `ReloadableX509...Manager` code. Although it looks like it might work on the surface, I think there's a danger that it will cause issues if the actual underlying server isn't expecting reloads.

I'm pretty sure in Tomcat there's some logic to ensure that rotated certs are only adopted for new connections, and connections that are already established continue to work as they are.

ukasus:
Hi @philwebb , Thanks for the feedback — that concern is valid, and I revisited the design with that in mind.

https://github.com/spring-projects/spring-boot/pull/50248
In above PR, I Simplified the approach to only a reloadable KeyManagerFactory (removed dual key/trust reload logic)
Reduced overall complexity and risk surface

I also explicitly validated the behavior:

1. Existing connections continue unaffected
2. Only new connections pick up rotated certs, similar to Apache Tomcat

Since the X509ExtendedKeyManager is only used during new TLS handshakes, there’s no impact on in-flight connections.

This should address the concern around unexpected reload behavior. Would appreciate a quick relook at the updated PR - https://github.com/spring-projects/spring-boot/pull/50248

PR #376, opened 2026-02-24
Add logging for gRPC server address and port on startup

Saksham-Prakash:
Closes #49049
This updates GrpcServerLifecycle startup logging to include both resolved listen address and resolved port in the same INFO log line


dsyer:
Isn't the port already part of the address that is logged (it is for me)?

Saksham-Prakash:
> Isn't the port already part of the address that is logged (it is for me)?

You’re right, thanks for pointing that out. In my environment getListenSockets() already includes the port (for example [127.0.0.1 (line 9090)](app://-/index.html#)), so the explicit port field is redundant in that case. I added it to make the resolved port explicit for #49049 and easier to grep, but I agree this can read as duplication. I’ll update it to avoid repeating the same information.

PR #381, opened 2026-03-03
Fix health auto-configuration condition timing

therepanic:
Fixes a timing issue. Currently, the logic in
`GrpcServerHealthAutoConfiguration.OnHealthIndicatorPathsCondition` is incorrect; it can never find the bean, since the condition itself is checked before the beans themselves are created. The tests passed because we were passing the bean itself, not the autoconfiguration with the bean.

Closes: spring-projects/spring-boot#49830


philwebb:
Thanks @therepanic! When the code was migrated to Spring Boot the condition changed to [use the `REGISTER_BEAN` phase.](https://github.com/spring-projects/spring-boot/blob/24a8053233f39223f1d4830dce6731e48b6f5776/module/spring-boot-grpc-server/src/main/java/org/springframework/boot/grpc/server/autoconfigure/health/GrpcServerHealthAutoConfiguration.java#L126)

I think that aligns things with the fix you've suggested here. We can double check when we get to looking at https://github.com/spring-projects/spring-boot/issues/49830

PR #384, opened 2026-03-05
Apply keepalive settings only when `enableKeepAlive` is true

therepanic:
This commit fixes an issue where we were applying keep-alive settings even when we had disabled the feature.

Closes: spring-projects/spring-boot#49829


philwebb:
Thanks @therepanic, but I'm going to close this one due to the code moving to Spring Boot and https://github.com/spring-projects/spring-boot/issues/49829#issuecomment-4218916993

PR #394, opened 2026-03-31
Add Spring Cloud discovery name resolver support

### LINKED ISSUE #341: Support Spring Cloud Discovery for NameResolver/NameResolverProvider
Considering the heavy usage of Spring Cloud Discovery in cloud-native projects, it would be great to support the integration of `spring-grpc` and `spring-cloud` for `NameResolver`.

The solution was implemented for https://github.com/grpc-ecosystem/grpc-spring, and I believe it can be entirely reused for spring-grpc.

https://github.com/yidongnan/grpc-spring-boot-starter/tree/master/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver

Hyainth:
## Summary
- add a vendor-neutral Spring Cloud discovery module for gRPC clients
- resolve `discovery:///service-name` targets through `DiscoveryClient`
- make the integration work end to end when the same module is also present on the server side

## What This Adds
- auto-configure a `NameResolverProvider` backed by Spring Cloud `DiscoveryClient`
- refresh resolved addresses when Spring Cloud publishes a `HeartbeatEvent`
- prefer the `gRPC_port` metadata entry and fall back to the service instance port
- publish `spring.grpc.server.port` as `gRPC_port` metadata through the Spring Cloud `Registration` abstraction when metadata is mutable
- register the discovery resolver before eager `GrpcChannelFactory` initialization so `discovery:///...` channels can be created during application startup
- keep the implementation generic to Spring Cloud instead of introducing a registry-specific starter

## Configuration
Client side:
- `spring.grpc.client.discovery.enabled=true`
- `spring.grpc.client.discovery.load-balancing-policy=round_robin`
- `spring.grpc.client.channels.<name>.address=discovery:///service-name`

Server side:
- `spring.grpc.server.port=9091`
- optional opt-out: `spring.grpc.server.discovery.publish-metadata=false`

## Verification
Automated:
- `JAVA_HOME=/usr/local/java/jdk-25.0.2.jdk/Contents/Home ./mvnw -Dmaven.repo.local=/Users/hyacinth/myRes/hurricane/project_dd/jdk25reposiory -pl spring-grpc-spring-cloud -am test`

Manual:
- validated locally with Spring Cloud Alibaba Nacos using only Spring Cloud Discovery integration plus this module, without a registry-specific gRPC discovery starter

Fixes #341

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,7 @@
 		<!-- Spring gRPC modules -->
 		<module>spring-grpc-build-dependencies</module>
 		<module>spring-grpc-core</module>
+		<module>spring-grpc-spring-cloud</module>
 		<module>spring-grpc-dependencies</module>
 		<module>spring-grpc-docs</module>
 <!--		<module>samples</module>-->
--- a/spring-grpc-build-dependencies/pom.xml
+++ b/spring-grpc-build-dependencies/pom.xml
@@ -54,6 +54,8 @@
 		<grpc-kotlin.version>1.5.0</grpc-kotlin.version>
 		<jspecify.version>1.0.0</jspecify.version>
 		<protobuf-java.version>4.34.0</protobuf-java.version>
+		<spring-boot.version>4.1.0-SNAPSHOT</spring-boot.version>
+		<spring-cloud-commons.version>5.0.1</spring-cloud-commons.version>
 		<spring-framework.version>7.0.6</spring-framework.version>
 		<spring-security.version>7.1.0-M2</spring-security.version>
 		<micrometer.version>1.17.0-M3</micrometer.version>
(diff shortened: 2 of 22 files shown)


Hyainth:
A bit more context on why this PR also includes the server-side metadata publication:

The client-side `DiscoveryClient` resolver alone is not enough to make `discovery:///...` channels work reliably for gRPC services registered through Spring Cloud service discovery.

In practice, service registration usually exposes the HTTP port as the instance port, while the gRPC server listens on `spring.grpc.server.port`. Without publishing that gRPC port separately, the resolver can only fall back to the regular instance port and the client ends up connecting to the wrong endpoint.

To keep the solution vendor-neutral, this PR does not introduce any Nacos-specific registration customizer. Instead, it uses the Spring Cloud `Registration` abstraction and only updates metadata when the underlying registration exposes a mutable metadata map. The behavior is also opt-out via `spring.grpc.server.discovery.publish-metadata=false`.

That keeps the integration generic to Spring Cloud while making the `spring-grpc-spring-cloud` module usable end to end for discovery-based gRPC clients.

dsyer:
We can't take a dependency on Spring Cloud (or Spring Boot) in this project. You will need to open a new PR against Spring Cloud.

PR #400, opened 2026-04-21
Add value alias to `@ImportGrpcClients`

### LINKED ISSUE #398: Add value alias to ImportGrpcClients annotation.
I think it would be helpful if `@ImportGrpcClients` had an alias for `value` that points to `types`. This would align with `ImportHttpServices` and allow users to write `@ImportGrpcClients(Foo.class)`.

therepanic:
Add a `value` attribute to `@ImportGrpcClients` as an alias for `types`, aligning with `@ImportHttpServices` and allowing the shorter form `@ImportGrpcClients(Foo.class)`.

Closes: gh-398

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/ImportGrpcClients.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/ImportGrpcClients.java
@@ -24,6 +24,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.springframework.context.annotation.Import;
+import org.springframework.core.annotation.AliasFor;
 
 /**
  * Annotation to create gRPC client beans.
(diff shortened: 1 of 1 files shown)


dsyer:
Seems unlikely to braeak anything. I'm still not sure what happens if the target is not specified.

therepanic:
So, I did a little research on this issue.

If we refer to the autoconfiguration logic we moved to Spring Boot, [we'll see that](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-grpc-client/src/main/java/org/springframework/boot/grpc/client/autoconfigure/PropertiesVirtualTargets.java#L41) with default [we'll use the default](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-grpc-client/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientProperties.java#L57) `"static://localhost:9090"`.

In general, I suggest we think about this. I can suggest the following:

We could force a custom target to be specified, whether as a property or an annotation parameter, and if it's not specified, we would return an error log. After that, we could remove the hardcoded code from Spring Boot's autoconfiguration.

This would be consistent with the logic behind, for example, `@EnableJpaRepositories`. After all, when we use this annotation, we don't connect to `jdbc:h2:mem:default` by default, and we see an error.

This would be the right thing to do, I think, and I can open an issue about it and we can figure out a solution.

therepanic:
This doesn't make sense now, because if the user forgets to specify their URL, they will see an error that the connection to `localhost:9090` has been broken.

Issue #401, opened 2026-04-21
Remove `localhost:9090` fallback for default gRPC channel


therepanic:
https://github.com/spring-projects/spring-grpc/pull/400#issuecomment-4289791912
> So, I did a little research on this issue.
> 
> If we refer to the autoconfiguration logic we moved to Spring Boot, [we'll see that](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-grpc-client/src/main/java/org/springframework/boot/grpc/client/autoconfigure/PropertiesVirtualTargets.java#L41) with default [we'll use the default](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-grpc-client/src/main/java/org/springframework/boot/grpc/client/autoconfigure/GrpcClientProperties.java#L57) "static://localhost:9090".
> 
> In general, I suggest we think about this. I can suggest the following:
> 
> We could force a custom target to be specified, whether as a property or an annotation parameter, and if it's not specified, we would return an error log. After that, we could remove the hardcoded code from Spring Boot's autoconfiguration.
> 
> This would be consistent with the logic behind, for example, @EnableJpaRepositories. After all, when we use this annotation, we don't connect to jdbc:h2:mem:default by default, and we see an error.
> 
> This would be the right thing to do, I think, and I can open an issue about it and we can figure out a solution.

If we're happy with this setup, I can submit a PR to the Spring Boot repository where I'll fix the autoconfiguration itself. After this, the user will be given a specific message that he should specify the address in the default channel.

dsyer:
As things stand if you use `@ImportGrpcClients` and not Spring Boot you get a default target of "default", so the `DefaultGrpcChannelFactory` will try to connect to that and get a `IllegalArgumentException` ("Could not find a NameResolverProvider for static://default"). The "static://" prefix is added by the `VirtualTargets.DEFAULT` and it was copied from the Spring gRPC community project, but it's not a thing in the Google grpc-java library. Nothing in spring-grpc-core actually understands it. We should probably make the default "localhost:9090" (with no schema added by the `VirtualTargets`) in this project? WDYT?

[...]

PR #404, opened 2026-04-22
Remove static:// prefix from VirtualTargets.DEFAULT

therepanic:
The static:// prefix is not a valid gRPC scheme and causes `IllegalArgumentException` when using `@ImportGrpcClients` without Spring Boot. Replace the prefix logic with a simple resolution of `"default"` to `"localhost:9090"`.

References: gh-401

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/client/VirtualTargets.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/client/VirtualTargets.java
@@ -16,20 +16,13 @@
 
 package org.springframework.grpc.client;
 
-import java.util.regex.Pattern;
-
 public interface VirtualTargets {
 
-	/** Regex to match the default authority pattern. */
-	Pattern AUTHORITY_PATTERN = Pattern.compile("([^:]+)(?::(\\d+))?");
-
-	/** Default VirtualTargets instance. */
-	VirtualTargets DEFAULT = path -> {
-		if (AUTHORITY_PATTERN.matcher(path).matches()) {
-			return "static://" + path;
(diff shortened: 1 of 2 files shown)


therepanic:
Btw, if we accept this PR, I can push another one to the Spring Boot repository where, for consistency, instead of:
```java
if ("default".equals(target)) {
    return clean(Channel.DEFAULT_TARGET); //<-- "static://localhost:9090" 
}
```

we'll do this:
```java
if ("default".equals(target)) {
    return VirtualTargets.DEFAULT.getTarget(target);
}
```

spring-grpc-core/src/main/java/org/springframework/grpc/client/VirtualTargets.java
```diff
@@ -16,20 +16,13 @@
...
-		if (AUTHORITY_PATTERN.matcher(path).matches()) {
-			return "static://" + path;
-		}
-		return path;
-	};
+	/**
+	 * Default VirtualTargets instance that resolves the logical name "default" to
+	 * "localhost:9090" and returns all other targets as-is.
+	 */
+	VirtualTargets DEFAULT = path -> "default".equals(path) ? "localhost:9090" : path;
```
therepanic, line 25: This will allow, when using the default VirtualTargets without Spring Boot, not to handle the case with `static://` and immediately substitute `default`.

dsyer:
I think that's the right outcome. Probably the right implementation (not sure). Maybe add a test for the `DefaultGrpcChannelFactory`?

therepanic:
Yeah, I added the test.

PR #410, opened 2026-05-16
Move `AUTHORIZATION_KEY` out of `GrpcSecurity` to avoid `ClassNotFoundException`

### LINKED ISSUE #409: BasicAuthenticationInterceptor causes ClassNotFoundException when spring-security-config is absent (client-only starter)
Bug Report
Summary
BasicAuthenticationInterceptor in spring-grpc-core imports GrpcSecurity, which extends AbstractConfiguredSecurityBuilder from spring-security-config. When using spring-grpc-client-spring-boot-starter (client-only), spring-security-config is not on the classpath, causing a ClassNotFoundException at runtime on every gRPC call.

Version
spring-grpc: 1.0.3 / Spring Boot: 4.0.6

Steps to Reproduce
Use spring-grpc-client-spring-boot-starter:1.0.3 (not the full spring-grpc-spring-boot-starter)
Configure BasicAuthenticationInterceptor as a @GlobalClientInterceptor
Make any outbound gRPC call

Actual Behaviour
java.lang.NoClassDefFoundError: org/springframework/security/config/annotation/AbstractConfiguredSecurityBuilder
    at BasicAuthenticationInterceptor$1.start(BasicAuthenticationInterceptor.java:53)
Caused by: java.lang.ClassNotFoundException: ...AbstractConfiguredSecurityBuilder

Root Cause
BasicAuthenticationInterceptor uses GrpcSecurity.AUTHORIZATION_KEY. GrpcSecurity is a server-side class extending AbstractConfiguredSecurityBuilder from spring-security-config. Loading the interceptor's anonymous class triggers loading of GrpcSecurity, which fails without spring-security-config on the classpath.

Suggested Fix
Move AUTHORIZATION_KEY out of GrpcSecurity into a shared constants class with no Spring Security dependencies, or define it inline in BasicAuthenticationInterceptor:

private static final Metadata.Key<String> AUTHORIZATION_KEY =
    Metadata.Key.of("Authorization", Metadata.ASCII_STRING_MARSHALLER);

therepanic:
The problem is that we don’t include the `spring-security` module in the client's auto-configuration, which causes a `ClassNotFoundException` when we try to use any of its constants. In our case, this is `GrpcSecurity#AUTHORIZATION_KEY`, and since client modules use it, we need to move it out. In our case, I moved it to a separate internal class called `GrpcMetadataKeys`.

Closes: gh-409

### DIFF
--- a/spring-grpc-core/src/main/java/org/springframework/grpc/internal/GrpcHeaders.java
+++ b/spring-grpc-core/src/main/java/org/springframework/grpc/internal/GrpcHeaders.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.grpc.internal;
+
+import io.grpc.Metadata;
+
+/**
+ * Common gRPC headers.
+ * <p>
+ * NOTE: Even though this class visibility is `public` it is intended for internal use
+ * only and not recommended for direct use.
+ *
+ * @author Andrey Litvitski
+ */
+public final class GrpcHeaders {
+
+	private GrpcHeaders() {
+	}
+
+	/**
+	 * A constant key used for storing and retrieving the "Authorization" header from gRPC
+	 * metadata. This key is used to handle authorization information in gRPC requests.
+	 *
+	 * <p>
+	 * The key is defined with the name "Authorization" and uses the ASCII string
+	 * marshaller for encoding and decoding the header value.
+	 * </p>
+	 */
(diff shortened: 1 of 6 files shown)


spring-grpc-core/src/main/java/org/springframework/grpc/internal/GrpcHeaders.java
```diff
@@ -0,0 +1,46 @@
...
+ * <p>
+ * NOTE: Even though this class visibility is `public` it is intended for internal use
+ * only and not recommended for direct use.
+ *
+ * @author Andrey Litvitski
+ */
+public final class GrpcMetadataKeys {
+
+	private GrpcMetadataKeys() {
+	}
```
therepanic, line 32: If anyone has any good suggestions for what to call this class, that would be great. Honestly, I haven't come up with anything for the current one either.

bkmuraliext: How about `GrpcHeaders`? It mirrors Spring's own `HttpHeaders` convention — a well-known precedent for a constants class holding header key names. Since gRPC's `Metadata` is the equivalent of HTTP headers, the analogy feels natural. It's short, discoverable, and leaves room to add other standard headers later without the name feeling off. `GrpcMetadataKeys` isn't bad, but "Keys" is slightly redundant since all the fields are keys by definition.

therepanic: Overall, it does look good - I was thinking the same thing myself. Thanks. I think it's definitely better than what we have now, though that’s debatable, but I still agree with you.

bkmuraliext (approved):
Thanks for the quick resolution!

dsyer:
Can we squash this down into a single commit please?

PR #431, opened 2026-09-16
Fix downdoc npx resolution when antora runs offline

rwinch:
The nested docs build spawned by @antora/collector-extension inherits npm_config_offline=true from the outer `npx --offline antora` process. Since downdoc is only resolved via npx (never installed locally in this build), that offline flag caused ENOTCACHED failures resolving downdoc from the registry. Pass --no-offline to the npx invocations so they resolve normally regardless of the inherited environment.


dsyer:
Applied t his change in b0fc2661. Thanks.

